PR #3, opened 2020-12-22
Cleanup api

turian:
So I've done a handful of cleanup things to the API. I'm still studying VCA and VCO so I didn't really touch them yet.

I submit this as a PR because there are several points of discussion, and maybe some of my changes you don't agree with or want to talk about.

Potentially controversially:
* I renamed the module to `ddspdrop` (from `ddrum`)
* I renamed the submodule to `modules` (from `drum_engine`)

Less controversially, the `parameters` are now `defaults`.

I've also changed some of the variable names around, but I'd love your suggestions. Standardizing on terminology (variable names) can really help the readers of the library.


examples/examples.py
```diff
@@ -0,0 +1,86 @@
+# # ddsp-drum examples
```
turian, line 1: I've move the "test.py" here. It makes more sense to run it through Jupyter notebook since it's more a demo than a formal test.

(Nonetheless, it's a nice way to informally test that the code still works as you expect.)

examples/examples.py
```diff
@@ -0,0 +1,86 @@
...
+
+# TODO: Why don't we add 's' here?
+dur = a + d + r + sustain_for
+
+# Envelope test
+adsr = ADSR(a, d, s, r, alpha)
+# TODO: Can you explain the envelope here?
+env = adsr(sustain_for)
+
+# Let's avoid cryptic variables names when possible
```
turian, line 40: I really want to avoid cryptic names, particularly because our audience might include people who don't have a DSP background.

src/ddspdrum/defaults.py
```diff
@@ -0,0 +1,7 @@
+"""
+Default parameters
+"""
+
```
turian, line 4: These are *defaults*, not fixed global parameters. I can imagine circumstances in which someone might want to have several synths running at once with different sample rates (e.g. to work with samples of multiple resolutions), so I have rewritten to allow this.

src/ddspdrum/module.py
```diff
@@ -0,0 +1,239 @@
...
+
+    def __init__(
+        self, sample_rate: int = SAMPLE_RATE, control_rate: int = CONTROL_RATE
+    ):
+        self.sample_rate = SAMPLE_RATE
+        self.control_rate = CONTROL_RATE
+
+    def control_to_sample_rate(self, control: np.array) -> np.array:
+        """
+        Resample a control signal to the sample rate
```
turian, line 27: @jorshi and @maxsolomonhenry I'd like your thoughts on this suggestion.

src/ddspdrum/module.py
```diff
@@ -0,0 +1,239 @@
...
+    def __init__(
+        self,
+        a: float = 0.25,
+        d: float = 0.25,
+        s: float = 0.5,
+        r: float = 0.5,
+        alpha: float = 3.0,
+        sample_rate: int = SAMPLE_RATE,
+        control_rate: int = CONTROL_RATE,
+    ):
```
turian, line 68: Okay, so one of the major changes is that we have to do away with this `get` and `set` nonsense. Having weird-side-effects makes thing harder to understand and also harder to port to neural networks.

We should be trying to write in a more *functional* way, without side-effects. It makes the code cleaner, and also neural networks are basically tensor functions.

src/ddspdrum/module.py
```diff
@@ -0,0 +1,239 @@
...
+        assert alpha >= 0
+        self.alpha = alpha
+
+        assert s >= 0 and s <= 1
+        self.s = s
+
+        assert a >= 0
+        self.a = a
+
+        # TODO: WHY? does it have to have an upper bound of 1?
```
turian, line 88: Please explain why `d` has an upper bound?

src/ddspdrum/module.py
```diff
@@ -0,0 +1,239 @@
...
+        assert d >= 0 and d <= 1
+        self.d = d
+
+        assert r >= 0
+        self.r = r
+
+    def __call__(self, duration):
```
turian, line 95: I am leaning towards standardizing on this idea that all `SynthModule`s should have a `__call__` method. I don't think we need `play` and `__call__`.

In torch, these will probably be renamed as `forward`.

src/ddspdrum/module.py
```diff
@@ -0,0 +1,239 @@
...
+        """
+        Play, for some duration in seconds.
+        """
+        assert duration >= 0
```
turian, line 99: This should actually be renamed to `sustain_duration`?

src/ddspdrum/module.py
```diff
@@ -0,0 +1,239 @@
...
+        return np.concatenate(
+            (
+                # Is this the DSP-way of thinking about it, (note on and note off)
+                # or can we change this just to self.attack, self.decay, etc.
+                self.note_on,
+                # This makes me believe that duration should maybe be a class value
+                # defined in __init__. How common is it to create a module and
+                # with fixed settings and use it several times with different
+                # duration?
+                self.sustain(duration),
```
turian, line 109: So one question we will need to solve is:
* what parameters go into `__init__`
* what parameters go into `forward`?

I am not sure whether this sustain duration should go into `forward` or `__init__`. As a rule of thumb, if it's the output from another module, then it goes into `forward` and otherwise it's a setting that goes in `__init__`.

src/ddspdrum/module.py
```diff
@@ -0,0 +1,239 @@
...
+        self.r = r
+
+    def __call__(self, duration):
+        """
+        Play, for some duration in seconds.
+        """
+        assert duration >= 0
+        return np.concatenate(
+            (
+                # Is this the DSP-way of thinking about it, (note on and note off)
```
turian, line 102: Would like your thoughts on this

src/ddspdrum/module.py
```diff
@@ -0,0 +1,239 @@
...
+            0, duration, int(round(duration * self.control_rate)), endpoint=False
+        )
+        return (t / duration) ** self.alpha
+
+    def __str__(self):
+        return (
+            f"ADRS(a={self.a}, d={self.d}, s={self.s}, r={self.r}, alpha={self.alpha})"
+        )
+
+
```
turian, line 154: So I didn't get a chance to scrutinize VCO and VCA yet but to reiterate:
* Don't use `get` and `set`, we don't want side effects.
* You can keep class parameters public. If you write things in a functional way (side-effect-free) then even if the user changes those values, forward() will still be correct.
* Parameters should go in `__init__` unless we expect that they are output by another module as input into this module, in which case they should go into `__call__` (later to be renamed `forward`). Don't worry to much about if we get this wrong now because it's easily fixed, particularly if we remove side-effects.

src/ddspdrum/module.py
```diff
@@ -0,0 +1,237 @@
...
+        TODO: One thing I worry about with all these casts to
+        ints back and forth between sample and control rate is
+        off-by-one errors.
+        I'm beginning to believe we should standardize as much
+        as possible on nsamples in most places instead of
+        float duration in seconds, and just convert to ncontrol
+        when needed..
+        """
+        # Right now it appears that all signals are 1d, but later
+        # we'll probably convert things to 2d: instance x signal
```
turian, line 37: So this is my current assumption about how I as a user would use this library, but tell me if you have other patterns in mind that we should have in the back of our minds in the roadmap:

* You define a particular synth architecture i.e. module layout (or maybe have several you are comparing side by side).
* You want to do some kind of training or inference (whether its synthesis or inverse-synthesis) in *mini-batches* not one patch at a time.
* If we standardize upon this idea that everything will be mini-batches, it's a bit simpler than the numpy style code-base where they constantly check if everything is one or two dimensions. The only real pain for the user is if they do want send one single patch through, in which they have to create a single row matrix.
* It's a little fiddly to convert from 1d to 2d and get everything right, so if we agree I think we fix this in the near future.
* Each module with then not have one set of parameters, but `batch_size` sets of parameters. This is a bit unusual for neural networks, that the network parameters depend upon the `batch_size`, but it's a pattern that seems to make the most sense to me given the few DDSP and inverse synth nets I've tried to build.

src/ddspdrum/module.py
jorshi: I've been thinking about the different rates for control signals and audio signals -- I like the idea and I think it makes sense for performance concerns. I'm also wondering if it is creating too much complexity at this point in time, especially considering we are resampling control signals up to audio signal rate when it comes time to apply them to signals. Should everything just run at audio rate?

src/ddspdrum/module.py
jorshi: Interesting! I was thinking about this a bit more as a traditional synth where you define your architecture (we can have pre-defined ones such as that Axxe drum machine), then you instantiate that and can set / update parameters on that one object and play sounds on that. 

But I think this is a bit different than what you are suggesting? So instead of having multiple instances of a synths within a batch, we would have one instance that can handle batch process -- And each module would need to support having mini-batches of parameters as well? We may have to decide on a simple parameter data structure to help keep this organized. If we go this route I would want to make sure that it is easy to use with a single parameter setting because that will be more intuitive for users coming from a music / synth background. I'm all for it if it will make the neural net stuff easier!

src/ddspdrum/module.py
jorshi: Sounds good -- I like the idea of have the __call__ method in the SynthModule base class

src/ddspdrum/module.py
jorshi: I'm not sure we need the note-on, note-off stuff here. To me that would be more for real-time use where you set a note-on which triggers the envelope and then holds the sustain (or wherever the envelope has reached) until a note-off message is received. If I understand correctly we have only fixed duration note lengths here. One way I've seen this handled in non-realtime stuff is to have a note-on length and a render length. So you process the envelope for the length of the note-on, and then move into the release portion of the envelope until the end of the render length.

src/ddspdrum/module.py
jorshi: Makes sense to me to set parameters in the constructor and then the output from another module gets passed into forward. It would be nice to have a way to set parameters after instantiation, I can see that be good for use after training, but we can cross that bridge later.

src/ddspdrum/defaults.py
jorshi: Nice, I think it is good to have the rates passed in during construction.

jorshi:
I'm all good with renaming to ddspdrum (I assume ddspdrum, not ddspdrop) and the rename to modules.

src/ddspdrum/module.py
maxsolomonhenry: I was wondering that myself. The benefit of a control rate comes when you're composing lots of control signals; LFOs and envelopes and whatnot. But given that this isn't really meant to run in realtime or anything, I see your point. Maybe let's leave it for now? Can always remove it later.

src/ddspdrum/module.py
maxsolomonhenry: My feeling, let's make sure it works in 1D before giving ourselves the headache. If it works slow we can then make it fast.

src/ddspdrum/module.py
maxsolomonhenry: It shouldn't be, my bad. Just must be non-negative.

src/ddspdrum/module.py
maxsolomonhenry: I was experimenting with that. The issue is that some calls will need arguments, or multiple arguments, and others won't. If there's a nice way of organizing this, I'd be happy to see it.

src/ddspdrum/module.py
maxsolomonhenry: Sure. What's important is that it's different from the "s" in ADSR, which is confusingly not a time-duration-value like the other parameters.

src/ddspdrum/module.py
maxsolomonhenry: We're going to want to separate note-ons and note-offs eventually for synth mapping. Much better to think about it this way than try to overengineer things, imo. It's two functions and will save us headaches down the line.

src/ddspdrum/module.py
maxsolomonhenry: At the moment, the envelope sets the global duration; that sets the length of the control signals, which ultimately sets the length of the audio signals. 

**Amp**-envelope parameters will determine the note-length (e.g., a release time of 0.5 adds 0.5 secs). Also the sustain amount.

Mind you, a frequency **cutoff**-envelope or a **pitch**-envelope can be shorter than the amp envelope, and just be padded with zeroes after to match the amp envelope. So we have to be careful about this. But when the amp envelope ends, the signal ends.

Maybe best to instantiate a signal with **amp** envelope ADR params plus sustain_duration. Then we have a fixed signal length and can allocate memory accordingly. What's tricky is that the amp envelope is not always the first step of the signal chain, in fact it's often the last.

src/ddspdrum/module.py
maxsolomonhenry: E.g., A VCO could be initiated with something like oscillator shape. But it should be **called** with pitch instructions (i.e., array of instantaneous frequency). That's how a synth works in practice anyway.

maxsolomonhenry: Feel free give the same treatment to the rest of the script (@property, etc).

examples/examples.py
maxsolomonhenry: The only names I could see as cryptic:

env --> envelope

t_cr --> timeline at control rate
t_sr --> timeline at sample rate

examples/examples.py
```diff
@@ -0,0 +1,86 @@
...
+a = 0.1
+d = 0.1
+s = 0.5
+r = 0.5
+alpha = 3
+sustain_for = 0.25
+
+f0 = 440
+
+# TODO: Why don't we add 's' here?
```
maxsolomonhenry, line 32: s is a volume value, not a duration value. One of the confusing things about the ADSR convention.

src/ddspdrum/module.py
maxsolomonhenry: Sure thing.

examples/examples.py
turian: Got it.

examples/examples.py
turian: yes. just more a note moving forward

src/ddspdrum/module.py
turian: @maxsolomonhenry what is the benefit? speed?

I do feel we should stop working in the domain of seconds and instead make nsamples (in sample rate domain) the primary unit of duration, pretty soon.

src/ddspdrum/module.py
turian: Okay, let's leave everything as 1D for now.

@jorshi I agree that multiple architectures is very important. One of the potentially cool things about what we're doing is architecture search. Traditional drum synths don't sound like real synths. I'm wondering how far we can push this towards natural drums by having complicated architecture. I'm not talking full on inception architecture stuff, but definitely more complex than usual. What's interesting about the DDSP approach is that in principle, if we can train neural networks on it, we can use the neural networks to program very complicated architectures to get the sounds we want.

src/ddspdrum/module.py
turian: Will fix

src/ddspdrum/module.py
turian: @maxsolomonhenry isn't it the case that __call__ can have args or multiple args or none? in which case just standardizing on __call__ or forward on stuff that takes args that outputs from other modules seems fine to me.

src/ddspdrum/module.py
turian: Got it

src/ddspdrum/module.py
turian: @maxsolomonhenry okay.

src/ddspdrum/module.py
```diff
@@ -0,0 +1,237 @@
...
+        applying the envelope curve and returning it in control rate.
+        """
+        t = np.linspace(
+            0, duration, int(round(duration * self.control_rate)), endpoint=False
+        )
+        return (t / duration) ** self.alpha
+
+    def __str__(self):
+        return (
+            f"ADRS(a={self.a}, d={self.d}, s={self.s}, r={self.r}, alpha={self.alpha})"
```
maxsolomonhenry, line 151: should be `f"ADSR`...

src/ddspdrum/module.py
turian: @maxsolomonhenry I don't exactly understand the concerns. Can you explain?

Which of these parameters is influenced by the output from another module which is input to this module? If we receive a freq cutoff or pitch envelope that is short then the rest of the parameters don't matter?

"Then we have a fixed signal length and can allocate memory accordingly." Does it make sense that the signal length should be---i don't want to say global---but at least something you define in advance before running the synth? Like sample rate and control rate?

Still don't fully grok this yet.

src/ddspdrum/module.py
turian: @maxsolomonhenry sounds good about how the pitch instructions being an input to map more naturally to how real synths work. This is where the "learn the synth params using NN" thing has a weirder/different pattern, if we also want to recover the pitch as part of inverse synthesis. In that case my proposal is we have a PitchGenerator module (in torch, a nn.Module) with a single learnable parameter pitch, that it just outputs, and is input to the VCO.

Thoughts?

src/ddspdrum/module.py
maxsolomonhenry: We have to keep the time components articulated in time values... we don't want them scaling with different sample rates.

src/ddspdrum/module.py
maxsolomonhenry: You're the pro. Works for me.

src/ddspdrum/module.py
maxsolomonhenry: Here's the idea. Let's assume: sample_rate = 44100, and control_rate= 1000, sustain_duration = 0.1s

Let's say you want a signal with a quick, snappy pitch attack. But you want the note to ring for a long time. 
You make an envelope for pitch--> 
```
pitch_env = ADSR(a=0.01, d=0.01,s=0,r=0)
```
This will be (0.01 + 0.01 + 0.1)*1000 = 120 samples long. But you want the sound to ring out, so you have an amp envelope with different parameters -->
```
amp_env = ADSR(a=0.01, d=0.25, s=0.75, r=0.5)
```
This will give an envelope that is (0.01 + 0.25 + 0.1 + 0.75) * 1000 = 1110 samples long.

This means we want a signal that is 1110 * 44100/1000 = 48951 samples at the audio rate.

src/ddspdrum/module.py
maxsolomonhenry: But we don't want to recover the pitch. We want to recover the pitch **parameters**, no? Now that you say this, it makes me think the VCO should have a base f0 frequency parameter, though.

src/ddspdrum/module.py
maxsolomonhenry: @turian Multiple subtractive synth architectures does not a real sound make. If you add up four motorcycles, do you get a car?

src/ddspdrum/module.py
turian: Got it.

src/ddspdrum/module.py
turian: @jorshi "It would be nice to have a way to set parameters after instantiation, I can see that be good for use after training, but we can cross that bridge later."

You just modify the member variable of the class, and then run the module again.

turian: @maxsolomonhenry So is what you're saying that we might have shorter pitch envelope and a shorter signal earlier in the chain, but the later modules might require longer signals and thus padding the earlier signal with zeros? I guess we can either a) zero pad or b) determine the final signal length at the beginning, with a parameter like `sample_rate` but instead `sample_duration`. a) sounds easier to start.

src/ddspdrum/module.py
turian: @maxsolomonhenry Yes we want to recover the pitch parameters.

src/ddspdrum/module.py
turian: @maxsolomonhenry not if you *add* them but if you chop them up and re-assemble them then maybe you do get a car. I'm not proposing linear regression on motorcycles. I'm proposing passing them through several non-linear operations.

PR #4, opened 2020-12-24
Filters

jorshi:
This will need to be updated when the changes to SynthModule are made. So will wait for that before merging. But wanted to post this here so we can discuss some ideas for filters. I've got three different types of filters here. One IIR and two different FIR. They each have their own strengths and weaknesses. The SVF is probably the best from a DSP perspective for what we are doing. But, it is an IIR filter. So may be problematic for the DDSP stuff. The FIR filter based on convolution with a windowed sinc function can make a really nice clean lowpass. It can be modulated (i think, still todo), but may not be very computationally efficient. And the simplest is a FIR moving average filter.


turian:
@jorshi I have merged with master and ran `black` and `isort`.

src/ddspdrum/module.py
```diff
@@ -235,3 +236,166 @@ def play(self):
...
+class SVF(SynthModule):
+    """
+    A State Variable Filter that can do lowpass, highpass, bandpass, and bandreject
+    filtering. Allows modulation of the cutoff frequency and an adjustable resonance
+    parameter. Can self-oscillate to make a sine / cosine wave.
+    """
+
+    def __init__(
+        self,
+        mode: str = "LPF",
```
turian, line 250: One thing I would like to avoid---as much as possible if it doesn't make us crazy---is having public modules that have categorical parameters ("LPF", "BPF", and "BSF"). My preference would be that we have a semiprivate `_SVF` module but public implementations are called `LPF` and `BPF` and `BSF`.

The idea is that backpropagating through a categorical variable is hard whereas, if there is an option to pickone of the three, we just have all three modules side by side, we use softmax to have an amplitude probability distribution go through the three (summing to one), and then potentially penalize the amplitude if its not in one-on two-off configuration.

src/ddspdrum/module.py
```diff
@@ -235,3 +236,166 @@ def play(self):
...
+        cutoff: float = 1000,
+        resonance: float = 0.707,
```
turian, line 252: What is the range here and what does it mean?

src/ddspdrum/module.py
```diff
@@ -235,3 +236,166 @@ def play(self):
...
+        self.mode = mode
+        self.cutoff = cutoff
+        self.resonance = resonance
+        self.self_oscillate = self_oscillate
+
+    def __call__(
+        self,
+        audio: np.ndarray,
+        cutoff_mod: np.ndarray = None,
+        cutoff_mod_amount: float = 0.0,
```
turian, line 267: What's the range here?

src/ddspdrum/module.py
```diff
@@ -235,3 +236,166 @@ def play(self):
...
+
+        h = np.zeros(2)
+        y = np.zeros_like(audio)
+
+        # Calculate initial coefficients
+        coeff0, coeff1, rho = self.calculate_coefficients(self.cutoff)
+
+        # Check if there is a filter cutoff envelope to apply
+        apply_modulation = False
+        if cutoff_mod is not None and cutoff_mod_amount != 0.0:
```
turian, line 281: I'm also slightly leery about mixing booleans and continuous values. Could we just use cutoff_mod_amount?

(Same reason being that if you are backprop'ing these parameters, the boolean will become a continuous in the range [0, 1])

src/ddspdrum/module.py
```diff
@@ -235,3 +236,166 @@ def play(self):
...
+        return coeff0, coeff1, rho
+
+
+class FIR(SynthModule):
+    """
+    A Finite Impulse Response filter
+    """
+
+    def __init__(
+        self,
```
turian, line 336: Same question here, what are the ranges and meanings of these?

src/ddspdrum/module.py
```diff
@@ -235,3 +236,166 @@ def play(self):
...
+    def __call__(self, audio: np.ndarray) -> np.ndarray:
+        """
+        Filter audio samples
+        TODO: Cutoff frequency modulation, if there is an efficient way to do it
+        """
+
+        impulse = self.windowed_sinc(self.cutoff, self.filter_length)
+        y = np.convolve(audio, impulse)
+        return y
+
```
turian, line 355: This can be a @property since all of it comes from the __init__ yeah?

src/ddspdrum/module.py
```diff
@@ -235,3 +236,166 @@ def play(self):
...
+    def windowed_sinc(self, cutoff: float, filter_length: float) -> np.ndarray:
+        """
+        Calculates the impulse response for FIR lowpass filter using the
+        windowed sinc function method
+        """
+
+        ir = np.zeros(filter_length + 1)
+        omega = 2 * np.pi * cutoff / self.sample_rate
+
+        for i in range(filter_length + 1):
```
turian, line 365: Is there a way to compute this without a for-loop?

src/ddspdrum/util.py
```diff
@@ -0,0 +1,30 @@
...
+    Convert decibel value to an amplitude between 0 and 1
+    """
+    return np.power(10, db / 20)
+
+
+def peak_gain_for_Q(Q: float):
+    """
+    Calculate the peak gain for a given filter quality factor.
+    """
+    # No gain added for quality factor less then 1/sqrt(2)
```
turian, line 26: We can leave this for now, but I'm also a bit concerned about how we will handle clamped values during backprop.

src/ddspdrum/module.py
```diff
@@ -235,3 +236,166 @@ def play(self):
...
+        super().__init__()
+        self.sample_rate = sample_rate
+        self.mode = mode
+        self.cutoff = cutoff
+        self.resonance = resonance
+        self.self_oscillate = self_oscillate
+
+    def __call__(
+        self,
+        audio: np.ndarray,
```
turian, line 265: Do we want to just call this `signal` everywhere? Or does it make sense to distinguish between `signal` and `audio`?

examples/filters.py
```diff
@@ -0,0 +1,172 @@
...
+#       format_name: light
+#       format_version: '1.5'
+#       jupytext_version: 1.8.0
+#   kernelspec:
+#     display_name: Python 3
+#     language: python
+#     name: python3
+# ---
+
+# # Filter Examples
```
turian, line 16: @jorshi Last suggestion. Let's put ALL the examples in one `examples.py` file, maybe with different sections.

Then it's easy to run code-coverage, which usually is easiest when you call only one file (`examples.py` in this case).

PR #7, opened 2020-12-27
Cleaning up VCO

turian:
@maxsolomonhenry there are a few things I need help with here, if it's a simple fix just push changes into this branch, or respond to comments if it's more discussion or there's something didactic for me.

### DIFF
--- a/src/ddspdrum/module.py
+++ b/src/ddspdrum/module.py
@@ -5,12 +5,10 @@ Synth modules.
                 - Convert operations to tensors, obvs.
 """
 
-import copy
-
 import numpy as np
 from scipy.signal import resample
 
-from ddspdrum.defaults import CONTROL_RATE, SAMPLE_RATE
+from ddspdrum.defaults import CONTROL_RATE, SAMPLE_RATE, EPSILON
 
 
 class SynthModule:
@@ -18,11 +16,10 @@ class SynthModule:
     Base class for synthesis modules. Mostly helper functions for the moment.
     """
 
-    def __init__(
-        self, sample_rate: int = SAMPLE_RATE, control_rate: int = CONTROL_RATE
-    ):
+    def __init__(self):
         self.sample_rate = SAMPLE_RATE
         self.control_rate = CONTROL_RATE
+        self.eps = EPSILON
 
     def control_to_sample_rate(self, control: np.array) -> np.array:
         """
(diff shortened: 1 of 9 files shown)


src/ddspdrum/module.py
```diff
@@ -173,31 +175,30 @@ def __init__(
...
-
-    def set_phase(self, val):
-        self.__phase = val % (2 * np.pi)
-
-    def play(self):
-        arg = self.to_arg(self.get_f0()) + self.get_phase()
-        self.set_phase(arg[-1])
+    def __call__(self, f0control: np.array, phase: float):
+        # TODO: Are there expected ranges on phase?
+        # Does it matter? Can we just assert preconditions here?
```
turian, line 180: Let's start enforcing preconditions on parameter values.

turian: It's important for making sure that randomly generated patches are sensible

src/ddspdrum/module.py
```diff
@@ -173,31 +175,30 @@ def __init__(
...
+        # What is arg and can we give it a better name?
+        arg = self.to_arg(clampedf0signal) + phase
+        # Why do we do a `set_phase` here at the end?
+        # If this is really some sort of recurrent state then maybe
+        # we return it and pass it back in?
+        #self.set_phase(arg[-1])
         return np.cos(arg)
 
-    def to_arg(self, f0):
+    def to_arg(self, f0control: np.array):
```
turian, line 197: @maxsolomonhenry can you please make this easier to understand? Maybe a better function name and docstring?

src/ddspdrum/module.py
```diff
@@ -173,31 +175,30 @@ def __init__(
...
+	    # because we lose the gradients from it, but let's worry
+        # about that later.
+        clampedf0control = np.clamp(f0control, 0, nyquist)
+
+        # What is arg and can we give it a better name?
+        arg = self.to_arg(clampedf0signal) + phase
+        # Why do we do a `set_phase` here at the end?
+        # If this is really some sort of recurrent state then maybe
+        # we return it and pass it back in?
+        #self.set_phase(arg[-1])
```
turian, line 194: @maxsolomonhenry this is what's really confusing me. Why do we have `self.set_phase(arg[-1])` at the end here?

turian: Why does it need to save its phase?

maxsolomonhenry: For the next output -- it has to start up at the previous phase value or else you'll get a click.

src/ddspdrum/module.py
maxsolomonhenry: Frequency is the derivative of phase. Phase is an integration of frequency. "argument" is what goes in the cosine, it's the phase.

src/ddspdrum/module.py
turian: simple asserts

PR #9, opened 2020-12-29
Demo drum class, plus cleanup stuff to util.py

maxsolomonhenry:
Drum class has a single method, and just does some stuff behind the scenes to make sure all the envelopes line up nicely, etc.

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -14,7 +14,7 @@ import librosa.display
 import matplotlib.pyplot as plt
 import numpy as np
 
-from ddspdrum.module import ADSR, VCA, VCO
+from ddspdrum.module import ADSR, VCA, VCO, Drum
 
 # -
 
@@ -92,6 +92,21 @@ plt.show()
 ipd.Audio(vca_out, rate=vca.sample_rate)
 # -
 
+# Alternately, you can just use the Drum class that composes all these modules together automatically.
 
+# +
+my_drum = Drum()
+
+boom = my_drum()
+
+X = librosa.stft(boom)
+Xdb = librosa.amplitude_to_db(abs(X))
+plt.figure(figsize=(5, 5))
+librosa.display.specshow(Xdb, sr=vca.sample_rate, x_axis="time", y_axis="hz")
+plt.ylim(0, 2000)
+plt.show()
+
+ipd.Audio(boom, rate=vca.sample_rate)
+# -
 
 
--- a/src/ddspdrum/module.py
+++ b/src/ddspdrum/module.py
@@ -8,8 +8,8 @@ Synth modules.
 import numpy as np
 from scipy.signal import resample
 
-from ddspdrum.defaults import CONTROL_RATE, SAMPLE_RATE, EPSILON
-
+from ddspdrum.defaults import CONTROL_RATE, SAMPLE_RATE 
+from ddspdrum.util import midi_to_hz, fix_length
 
 class SynthModule:
     """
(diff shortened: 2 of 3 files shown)


src/ddspdrum/module.py
```diff
@@ -240,14 +216,15 @@ def __call__(self, mod_signal: np.array, phase: float) -> np.array:
         assert (mod_signal >= 0).all() and (mod_signal <= 1).all()
 
         control_as_midi = self.mod_depth * mod_signal + self.midi_f0
-        control_as_frequency = self.midi_to_hz(control_as_midi)
+        control_as_frequency = midi_to_hz(control_as_midi)
```
maxsolomonhenry, line 219: now coming from util.py

src/ddspdrum/module.py
```diff
@@ -268,10 +245,46 @@ def __call__(self, envelopecontrol: np.array, audiosample: np.array):
         envelopecontrol = np.clip(envelopecontrol, 0, 1)
         audiosample = np.clip(audiosample, -1, 1)
         amp = self.control_to_sample_rate(envelopecontrol)
-        signal = self.fix_length(audiosample, len(amp))
+        signal = fix_length(audiosample, len(amp))
         return amp * signal
 
 
+class Drum:
```
maxsolomonhenry, line 252: will fold the filters into this eventually when I figure out the differentiability aspect of them.

examples/examples.py
```diff
@@ -92,6 +92,21 @@
 ipd.Audio(vca_out, rate=vca.sample_rate)
 # -
 
+# Alternately, you can just use the Drum class that composes all these modules together automatically.
 
+# +
+my_drum = Drum()
+
+boom = my_drum()
```
turian, line 100: let's just call this signal or X ? (and the STFT is Z?)

src/ddspdrum/module.py
```diff
@@ -108,7 +84,7 @@ def __init__(
         assert r >= 0
         self.r = r
 
-    def __call__(self, sustain_duration=0):
+    def __call__(self, sustain_duration: float = 0):
```
turian, line 87: Why do we need this default? Don't we prefer that people specify this value rather than giving a weird default?

src/ddspdrum/module.py
```diff
@@ -268,10 +245,47 @@ def __call__(self, envelopecontrol: np.array, audiosample: np.array):
...
 
 
+class Drum:
+    """
+    A package of modules that makes one drum hit. 
+    TODO:   - clicks when the envelopes don't match up.
+    """
+
+    def __init__(
+        self,
```
turian, line 259: I think it would be cleaner and safer to construct this by passing in:

```
pitch_adsr: ADSR, amp_adsr: ADSR, vco: VCO, vca: VCA
```

Otherwise it's easy to fuck up the parameters

examples/examples.py
maxsolomonhenry: ok. changed it to "drum_out" to match the formatting so far.

src/ddspdrum/module.py
maxsolomonhenry: simply put, an ADSR with a 0 sec sustain (i.e., no sustain) is an ADR envelope, which is also common.

maxsolomonhenry: (especially for drums.)

src/ddspdrum/module.py
maxsolomonhenry: I agree this is weird, but the idea is the drum module does all the work for you, otherwise you might as well build it from scratch, no?

maxsolomonhenry: nm, it's all good I did it.

PR #10, opened 2020-12-31
Osc update -- adding square / saw

maxsolomonhenry:
Pretty straight forward. Sounds rad too. Paper is cited in the code.

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -14,7 +14,7 @@ import librosa.display
 import matplotlib.pyplot as plt
 import numpy as np
 
-from ddspdrum.module import ADSR, VCA, VCO, Drum
+from ddspdrum.module import ADSR, VCA, VCO, SquareSaw, Drum
 
 # -
 
@@ -70,6 +70,28 @@ plt.show()
 ipd.Audio(vco_out, rate=vco.sample_rate)
 # -
 
+# Check this out, it's a square / saw oscillator. Use the shape parameter to interpolate between a square wave (shape = 0) and a sawtooth wave (shape = 1).
+
+# +
+# SquareSaw test: shape 0 --> square, 1 --> saw.
+
+shape = 0
+
+midi_f0 = 55
+sqs = SquareSaw(shape=shape, midi_f0=midi_f0, mod_depth=6)
+sqs_out = sqs(envelope, phase=0)
+
+# +
+X = librosa.stft(sqs_out)
+Xdb = librosa.amplitude_to_db(abs(X))
+plt.figure(figsize=(5, 5))
+librosa.display.specshow(Xdb, sr=vco.sample_rate, x_axis="time", y_axis="hz")
+plt.ylim(0, 5000)
+plt.show()
+
+ipd.Audio(sqs_out, rate=vco.sample_rate)
+# -
+
 # Notice that this sound is rather clicky. We'll add an envelope to the amplitude to smooth it out.
 
 # VCA test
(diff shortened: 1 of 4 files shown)


src/ddspdrum/module.py
```diff
@@ -239,6 +239,38 @@ def make_argument(self, control_as_frequency: np.array):
         up_sampled = self.control_to_sample_rate(control_as_frequency)
         return np.cumsum(2 * np.pi * up_sampled / SAMPLE_RATE)
 
+    def oscillator(self, argument):
+        return np.cos(argument)
+
+
+class SquareSaw(VCO):
```
maxsolomonhenry, line 246: Inherits from VCO (might be a better way to do this, but it's pretty smooth).

examples/examples.py
```diff
@@ -95,18 +117,47 @@
...
 plt.figure(figsize=(5, 5))
 librosa.display.specshow(Xdb, sr=vca.sample_rate, x_axis="time", y_axis="hz")
 plt.ylim(0, 2000)
 plt.show()
 
-ipd.Audio(boom, rate=vca.sample_rate)
+ipd.Audio(drum_out, rate=vca.sample_rate)
 # -
+# You can also use the **SquareSaw oscillator** in the drum module.
+
```
maxsolomonhenry, line 139: Showing off the new oscillator.

src/ddspdrum/module.py
```diff
@@ -229,7 +229,7 @@ def __call__(self, mod_signal: np.array, phase: float = 0) -> np.array:
         cosine_argument = self.make_argument(control_as_frequency) + phase
 
         self.phase = cosine_argument[-1]
-        return np.cos(cosine_argument)
+        return self.oscillator(cosine_argument)
```
maxsolomonhenry, line 232: The idea being, this can be inherited and overrided with difference oscillator types. Would it be cleaner to have a VCO base class and a SINE specific class to inherit? Whatever works...

examples/examples.py
```diff
@@ -95,18 +117,47 @@
...
+my_drum = Drum(
+    pitch_adsr=ADSR(0.25, 0.25, 0.25, 0.25, alpha=3),
+    amp_adsr=ADSR(0.25, 0.25, 0.25, 0.25),
+    vco=SquareSaw(shape=0, midi_f0=24, mod_depth=12),
+    sustain_duration=1
+    )
+
+drum_out = my_drum()
+
+X = librosa.stft(drum_out)
```
turian, line 150: How many times are we doing this? Maybe we should turn it into a function?

src/ddspdrum/module.py
turian: I'm into this

src/ddspdrum/module.py
```diff
@@ -239,6 +239,38 @@ def make_argument(self, control_as_frequency: np.array):
         up_sampled = self.control_to_sample_rate(control_as_frequency)
         return np.cumsum(2 * np.pi * up_sampled / SAMPLE_RATE)
 
+    def oscillator(self, argument):
```
turian, line 242: I'm not sure but maybe we make VCO into an abstract class and we have a second subclass that is a SineVCO with this oscillator?

I am also feeling like oscillator should be a param in VCO __init__ but not 100% on it

src/ddspdrum/module.py
turian: Let's call it SquareSawVCO? I'm not so into Polish variable names but I think it makes sense

src/ddspdrum/module.py
```diff
@@ -239,6 +239,38 @@ def make_argument(self, control_as_frequency: np.array):
...
+        mod_depth: float = 50, 
+        phase: float = 0    
+    ):
+        super().__init__(midi_f0, mod_depth, phase)
+        assert 0 <= shape <= 1 
+        self.shape = shape
+
+    def oscillator(self, argument):
+        k = self.get_k()
+        square = np.tanh(np.pi * k * np.sin(argument)/2)
```
turian, line 268: I'd be very curious to see this plotted in example.py but not essential by any means

turian: The square fn

src/ddspdrum/module.py
maxsolomonhenry: hah. I mean, if you like. It's getting a bit long as a name at that point. I'll make the change for now.

examples/examples.py
maxsolomonhenry: meh.

src/ddspdrum/module.py
maxsolomonhenry: ok let's do that then. not 100% on having oscillator type as a param though. how would that work, structurally?

src/ddspdrum/module.py
maxsolomonhenry: it's just a clipped sin, basically. kindof a clever paper. but it's guts stuff, not sure if it needs to be upfront.

src/ddspdrum/module.py
```diff
@@ -239,6 +239,38 @@ def make_argument(self, control_as_frequency: np.array):
...
+        return (1 - self.shape/2) * square * (1 + self.shape * np.cos(argument))
+
```
turian, line 290: This stuff that is completely dependent on __init__ params can be a @property

turian (approved):

PR #11, opened 2021-01-05
A bit of cleanup and refactoring

turian:
Mainly, as discussed before, sample_rate and control_rate are no longer globals.

### DIFF
--- a/src/ddspdrum/module.py
+++ b/src/ddspdrum/module.py
@@ -17,9 +17,9 @@ class SynthModule:
     Base class for synthesis modules.
     """
 
-    def __init__(self):
-        self.sample_rate = SAMPLE_RATE
-        self.control_rate = CONTROL_RATE
+    def __init__(self, sample_rate: int = SAMPLE_RATE, control_rate: int = CONTROL_RATE):
+        self.sample_rate = sample_rate
+        self.control_rate = control_rate
 
     def control_to_sample_rate(self, control: np.array) -> np.array:
         """
@@ -57,6 +57,7 @@ class ADSR(SynthModule):
         s: float = 0.5,
         r: float = 0.5,
         alpha: float = 3.0,
+        sample_rate: int = SAMPLE_RATE, control_rate: int = CONTROL_RATE
     ):
         """
         Parameters
@@ -70,8 +71,8 @@ class ADSR(SynthModule):
         alpha               :   envelope curve, >= 0. 1 is linear, >1 is
                                 exponential.
         """
+        super().__init__(sample_rate=sample_rate, control_rate=control_rate)
 
-        super().__init__()
         assert alpha >= 0
         self.alpha = alpha
 
@@ -186,8 +187,9 @@ class VCO(SynthModule):
     >>> two_8ve_chirp = myVCO(np.linspace(0, 1, 1000, endpoint=False))
     """
 
-    def __init__(self, midi_f0: float = 10, mod_depth: float = 50, phase: float = 0):
-        super().__init__()
+    def __init__(self, midi_f0: float = 10, mod_depth: float = 50, phase: float = 0,         sample_rate: int = SAMPLE_RATE, control_rate: int = CONTROL_RATE
+):
+        super().__init__(sample_rate=sample_rate, control_rate=control_rate)
 
         assert 0 <= midi_f0 <= 127
         self.midi_f0 = midi_f0
@@ -292,8 +294,10 @@ class VCA(SynthModule):
     Voltage controlled amplifier.
     """
 
-    def __init__(self):
-        super().__init__()
+    def __init__(self,         sample_rate: int = SAMPLE_RATE, control_rate: int = CONTROL_RATE
+):
+        super().__init__(sample_rate=sample_rate, control_rate=control_rate)
+
         self.__envelope = np.array([])
         self.__audio = np.array([])
 
@@ -352,10 +356,10 @@ class SVF(SynthModule):
         cutoff: float = 1000,
         resonance: float = 0.707,
         self_oscillate: bool = False,
-        sample_rate: int = SAMPLE_RATE,
+        sample_rate: int = SAMPLE_RATE, control_rate: int = CONTROL_RATE
     ):
-        super().__init__()
-        self.sample_rate = sample_rate
+        super().__init__(sample_rate=sample_rate, control_rate=control_rate)
+
         self.mode = mode
         self.cutoff = cutoff
         self.resonance = resonance
@@ -437,10 +441,10 @@ class FIR(SynthModule):
         self,
         cutoff: float = 1000,
         filter_length: int = 512,
-        sample_rate: int = SAMPLE_RATE,
+        sample_rate: int = SAMPLE_RATE, control_rate: int = CONTROL_RATE
     ):
-        super().__init__()
-        self.sample_rate = sample_rate
+        super().__init__(sample_rate=sample_rate, control_rate=control_rate)
+
         self.filter_length = filter_length
         self.cutoff = cutoff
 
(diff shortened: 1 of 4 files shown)


src/ddspdrum/module.py
```diff
@@ -380,6 +419,7 @@ def __call__(
 
             # If there is a cutoff modulation envelope, update coefficients
             if apply_modulation:
+                # BUG: Cutoff variable never used?
```
turian, line 422: @jorshi just curious about this

src/ddspdrum/module.py
```diff
@@ -163,7 +169,8 @@ def __str__(self):
 
 
 class VCO(SynthModule):
-    """Voltage controlled oscillator.
+    """
```
maxsolomonhenry, line 172: pep8 has a one-liner up top:
`
"""This is what the function does.

   Details go here. Args etc.
`

src/ddspdrum/module.py
```diff
@@ -186,8 +193,15 @@ class VCO(SynthModule):
     >>> two_8ve_chirp = myVCO(np.linspace(0, 1, 1000, endpoint=False))
     """
 
-    def __init__(self, midi_f0: float = 10, mod_depth: float = 50, phase: float = 0):
-        super().__init__()
+    def __init__(
+        self,
+        midi_f0: float = 10,
+        mod_depth: float = 50,
```
turian, line 199: @maxsolomonhenry what is the upper bound on mod_depth? Surely it can't be infinity?

src/ddspdrum/module.py
```diff
@@ -17,9 +19,11 @@ class SynthModule:
     Base class for synthesis modules.
     """
 
-    def __init__(self):
-        self.sample_rate = SAMPLE_RATE
-        self.control_rate = CONTROL_RATE
```
maxsolomonhenry, line 22: I'm still really not sure why you want to make your life so complicated. But if it makes you happy I'm fine with it.

src/ddspdrum/module.py
```diff
@@ -198,7 +212,8 @@ def __init__(self, midi_f0: float = 10, mod_depth: float = 50, phase: float = 0)
         self.phase = phase
 
     def __call__(self, mod_signal: np.array, phase: float = 0) -> np.array:
-        """Generates audio signal from control-rate mod.
+        """
+        Generates audio signal from control-rate mod.
```
maxsolomonhenry, line 216: See above about pep8, etc. etc.

src/ddspdrum/module.py
```diff
@@ -305,20 +326,37 @@ def __call__(self, envelopecontrol: np.array, audiosample: np.array):
         return amp * signal
 
 
-class Drum:
+class Synth:
+    """
+    An abstract class for a modular synth, ensuring that all modules
+    have the same sample and control rate.
+    """
```
maxsolomonhenry, line 334: Again, if they all have the same abstract base class I don't see why this is necessary. But again, fine.

src/ddspdrum/module.py
turian: Global variables are easy ways to shoot yourself in the foot down the road, and to create hard to find bug.

src/ddspdrum/module.py
```diff
@@ -305,20 +326,37 @@ def __call__(self, envelopecontrol: np.array, audiosample: np.array):
...
+        sustain_duration: float,
         pitch_adsr: ADSR = ADSR(),
         amp_adsr: ADSR = ADSR(),
         vco: VCO = VCO(),
         vca: VCA = VCA(),
-        sustain_duration: float = 0,
     ):
+        super().__init__(modules=[pitch_adsr, amp_adsr, vco, vca])
+        # TODO: Should we make this strictly greater than 0.0?
+        assert sustain_duration >= 0
```
maxsolomonhenry, line 358: Why would it ever be negative?

src/ddspdrum/module.py
turian: If you want. My preference is we use
```
"""one-liner"""
```

and
```
"""
line1
line2
multiline comment
"""
```

style

src/ddspdrum/module.py
maxsolomonhenry: 0 sustain is fine, for now. We might want to rework the adsr because it's note quite like midi atm, and more setup for drums. can discuss this later.

src/ddspdrum/module.py
turian: @maxsolomonhenry both are valid PEP8 according to `black` and `flake8`

src/ddspdrum/module.py
turian: @maxsolomonhenry the way it was implemented before this PR did not make it simple/safe to have a single Jupyter notebook with two Drum instances with different rates.

src/ddspdrum/module.py
maxsolomonhenry: Works for me.

src/ddspdrum/module.py
turian: Because it was generated by a neural network? And because these assertions are a useful way for now for documenting way the expected preconditions are for these values?

src/ddspdrum/module.py
maxsolomonhenry: Currently mod depth is in semitones. So the upper bound is 127, assuming osc midi_f0 is 0. 

Otherwise upper bound is (127 - osc midi_f0)

src/ddspdrum/module.py
turian: (Because it was generated by a neural network unfamiliar with the API, since we don't have a programmatic way to query the parameter ranges)

src/ddspdrum/module.py
maxsolomonhenry: Cool.

src/ddspdrum/module.py
maxsolomonhenry: ¯\_(ツ)_/¯ 

Ok. I trust your experience.

src/ddspdrum/module.py
turian: Okay great.

src/ddspdrum/module.py
maxsolomonhenry: Ok. Well, either way a 0 is fine. So it can stay >=0

src/ddspdrum/module.py
turian: @maxsolomonhenry what why is mod_depth >= 0? Why can't you also modulate it down?

maxsolomonhenry: I don't know if you want to get into this now, or just make a note of it.

maxsolomonhenry: I suppose you could, sure. Would have another bound, obviously. -- but then we've have to update the SquareSaw Oscillator... k calculation.

src/ddspdrum/module.py
maxsolomonhenry: Wanna strip that comment?

src/ddspdrum/module.py
turian: I don't think it needs to be changed, it looks fine?

```
        # What does k mean here? Can we give it a better name?
        f0 = midi_to_hz(self.midi_f0 + self.mod_depth)
        return 12000 / (f0 * np.log10(f0))
```

Or is mod_depth just non-negative by convention?

What is `k` anyway? Is there a better name for it?

src/ddspdrum/module.py
```diff
@@ -182,7 +182,7 @@ class VCO(SynthModule):
     ----------
 
     midi_f0 (flt)       :       pitch value in 'midi' (69 = 440Hz).
-    mod_depth (flt)     :       depth of the pitch modulation in semitones; 0 means none.
+    mod_depth (flt)     :       depth of the pitch modulation in semitones.
 
     TODO:   - more than just cosine.
```
maxsolomonhenry, line 187: We can probably kill this. I'll grab it if you don't.

turian: Removed

src/ddspdrum/module.py
```diff
@@ -305,20 +326,36 @@ def __call__(self, envelopecontrol: np.array, audiosample: np.array):
...
         pitch_adsr: ADSR = ADSR(),
         amp_adsr: ADSR = ADSR(),
         vco: VCO = VCO(),
         vca: VCA = VCA(),
-        sustain_duration: float = 0,
     ):
+        super().__init__(modules=[pitch_adsr, amp_adsr, vco, vca])
+        assert sustain_duration >= 0
 
+        # TODO: Do we want the same sustain_duration for both ADSRs?
```
maxsolomonhenry, line 358: That's the convention for triggering a note event, yes.

turian: Updated comment

PR #12, opened 2021-01-05
Filter cleanup

jorshi:
Cleaning up filter stuff from Joseph's comments on previous pull request


maxsolomonhenry (approved):
Works for me.

PR #16, opened 2021-01-05
Noise gen plus multiple oscillators

maxsolomonhenry:
Added simple noise generator to the Drum class, plus the ability to accept an arbitrary number of oscillators.

(We may in fact want to fix this to two oscillators, with a mix parameter between them.)

(...)

(Still, sounds pretty cool. Check the examples.)

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -112,6 +112,22 @@ stft_plot(drum_out)
 
 ipd.Audio(drum_out, rate=vca.sample_rate)
 # -
+# Use the `noise_ratio` parameter to add noise.
+
+# +
+my_drum = Drum(
+    pitch_adsr=ADSR(0.25, 0.25, 0.25, 0.25, alpha=3),
+    amp_adsr=ADSR(0.25, 0.25, 0.25, 0.25),
+    vco=SineVCO(midi_f0=55, mod_depth=6),
+    sustain_duration=1,
+    noise_ratio=0.5
+)
+
+drum_out = my_drum()
+stft_plot(drum_out)
+ipd.Audio(drum_out, rate=vca.sample_rate)
+# -
+
 # You can also use the **SquareSawVCO oscillator** in the drum module.
 
 
@@ -121,6 +137,24 @@ my_drum = Drum(
     amp_adsr=ADSR(0.25, 0.25, 0.25, 0.25),
     vco=SquareSawVCO(shape=0, midi_f0=24, mod_depth=12),
     sustain_duration=1,
+)
+
+drum_out = my_drum()
+stft_plot(drum_out)
+ipd.Audio(drum_out, rate=vca.sample_rate)
+# -
+
+# Additionally, the Drum class can take any number of oscillators.
+
+# +
+my_drum = Drum(
+    pitch_adsr=ADSR(0.25, 0.25, 0.25, 0.25, alpha=3),
+    amp_adsr=ADSR(0.25, 0.25, 0.25, 0.25),
+    vco=[
+        SquareSawVCO(shape=0, midi_f0=23.95, mod_depth=12),
+        SquareSawVCO(shape=0, midi_f0=24.05, mod_depth=12)
+    ],
+    sustain_duration=1
 )
 
 drum_out = my_drum()
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -97,7 +97,7 @@ noisey_out = noiser(sqs_out)
 
 time_plot(noisey_out)
 stft_plot(noisey_out)
-ipd.Audio(noisey_out, rate=vca.sample_rate)
+ipd.Audio(noisey_out, rate=sqs.sample_rate)
 
 # Notice that this sound is rather clicky. We'll add an envelope to the amplitude to smooth it out.
 
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -15,8 +15,8 @@ import librosa.display
 import matplotlib.pyplot as plt
 import numpy as np
 
-from ddspdrum.module import ADSR, VCA, Drum, SineVCO, SquareSawVCO, NoiseModule
 from ddspdrum.defaults import SAMPLE_RATE
+from ddspdrum.module import ADSR, VCA, Drum, NoiseModule, SineVCO, SquareSawVCO
 
 # -
 
@@ -132,11 +132,11 @@ ipd.Audio(drum_out, rate=vca.sample_rate)
 my_drum = Drum(
     pitch_adsr=ADSR(0.25, 0.25, 0.25, 0.25, alpha=3),
     amp_adsr=ADSR(0.25, 0.25, 0.25, 0.25),
-    vco_1 = SquareSawVCO(shape=0, midi_f0=23.95, mod_depth=12),
-    vco_2 = SquareSawVCO(shape=0, midi_f0=24.05, mod_depth=12),
-    vco_1_ratio = 0.5,
-    noise_module = NoiseModule(ratio=0.1),
-    sustain_duration=1
+    vco_1=SquareSawVCO(shape=0, midi_f0=23.95, mod_depth=12),
+    vco_2=SquareSawVCO(shape=0, midi_f0=24.05, mod_depth=12),
+    vco_1_ratio=0.5,
+    noise_module=NoiseModule(ratio=0.1),
+    sustain_duration=1,
 )
 
 drum_out = my_drum()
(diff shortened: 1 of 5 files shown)


examples/examples.py
```diff
@@ -128,6 +144,24 @@ def stft_plot(signal, sample_rate=SAMPLE_RATE):
...
+# Additionally, the Drum class can take any number of oscillators.
+
+# +
+my_drum = Drum(
+    pitch_adsr=ADSR(0.25, 0.25, 0.25, 0.25, alpha=3),
+    amp_adsr=ADSR(0.25, 0.25, 0.25, 0.25),
+    vco=[
+        SquareSawVCO(shape=0, midi_f0=23.95, mod_depth=12),
+        SquareSawVCO(shape=0, midi_f0=24.05, mod_depth=12)
+    ],
```
maxsolomonhenry, line 156: As I say, we may just want to fix these to two oscillators. And we might want a mix knob between the two of them. At the moment it just does an average of all oscillators.

src/ddspdrum/module.py
```diff
@@ -319,15 +319,12 @@ def __init__(
...
-        audiosample = np.clip(audiosample, -1, 1)
-        amp = self.control_to_sample_rate(envelopecontrol)
-        signal = fix_length(audiosample, len(amp))
-        return amp * signal
+    def __call__(self, control_in: np.array, audio_in: np.array):
+        control_in = np.clip(control_in, 0, 1)
+        audio_in = np.clip(audio_in, -1, 1)
+        amp = self.control_to_sample_rate(control_in)
+        audio_in = fix_length(audio_in, len(amp))
+        return amp * audio_in
```
maxsolomonhenry, line 327: Readable variable names

src/ddspdrum/module.py
```diff
@@ -353,10 +350,14 @@ def __init__(
...
         amp_adsr: ADSR = ADSR(),
-        vco: VCO = VCO(),
+        vco: Union[VCO, List[VCO]] = VCO(),
         vca: VCA = VCA(),
+        noise_ratio: float = 0
     ):
-        super().__init__(modules=[pitch_adsr, amp_adsr, vco, vca])
+        if type(vco) is not list:
+            vco = [vco]
+
```
maxsolomonhenry, line 359: This feels messy. Not sure if this is the best way to do this (i.e., accept arbitrary number of oscillators).

src/ddspdrum/module.py
```diff
@@ -365,12 +366,27 @@ def __init__(
...
+    def play_vco_bank(vco_bank, pitch_envelope):
+        audio_out = 0
+        for vco in vco_bank:
+            audio_out += vco(pitch_envelope)
+        return audio_out / len(vco_bank)
+
+    @staticmethod
+    def add_noise(audio_in, ratio):
+        return (1 - ratio) * audio_in + ratio * np.random.rand(len(audio_in))
+
```
maxsolomonhenry, line 389: This too is a bit messy. Might be a better way to refactor.

src/ddspdrum/module.py
```diff
@@ -353,10 +350,14 @@ def __init__(
         sustain_duration: float,
         pitch_adsr: ADSR = ADSR(),
         amp_adsr: ADSR = ADSR(),
-        vco: VCO = VCO(),
+        vco: Union[VCO, List[VCO]] = VCO(),
```
turian, line 353: Can we just force it to be a list of VCOs and call the variable vcos? This Union this is kinda weird.

src/ddspdrum/module.py
```diff
@@ -353,10 +350,14 @@ def __init__(
...
         vca: VCA = VCA(),
+        noise_ratio: float = 0
     ):
-        super().__init__(modules=[pitch_adsr, amp_adsr, vco, vca])
+        if type(vco) is not list:
```
turian, line 357: Then we can remove this, which is not considered very pythonic

src/ddspdrum/module.py
turian: Force it to be a Tuple (not List) instead of Union. Each should have a weight between 0 and 1. Don't enforce that they sum to one.

I still think at some point soon we'll want to make a decision upon a default specific synth structure, i.e. number of oscillators, since learning through a variable structure is hard. We could keep the generic var-osc thing, but I think the end-user API we should just pick a sensible default.

src/ddspdrum/module.py
```diff
@@ -365,12 +366,27 @@ def __init__(
         self.amp_envelope = amp_adsr(sustain_duration)
         self.vco = vco
         self.vca = vca
+        self.noise_ratio = noise_ratio
```
turian, line 369: Should we call this vco_noise_ratio?

turian:
Generally looks pretty good to me, but see comments

src/ddspdrum/module.py
maxsolomonhenry: Ok

src/ddspdrum/module.py
maxsolomonhenry: The idea is that you could just feed it a single VCO and it would work fine. It's an interface choice. But as you say, might be better to work with a fixed architecture anyway. Thoughts?

src/ddspdrum/module.py
maxsolomonhenry: Yeah it's gross.

src/ddspdrum/module.py
maxsolomonhenry: It's not a vco though, it's part of the drum module.

src/ddspdrum/module.py
maxsolomonhenry: Ok then, shall we go with two oscillators?

src/ddspdrum/module.py
turian: I don't really have good insight on how many oscillators is enough :) Your call with @jorshi

src/ddspdrum/module.py
turian: Sure

src/ddspdrum/module.py
turian: Oh I just meant it's the noise ratio applied to the vco bank. Okay

src/ddspdrum/module.py
jorshi: Fixing with two for now sounds good. When we were starting this project I imagined that we would have a collection of building block dsp modules, and then some preset architectures based on existing drum machines. Like an Arp Axxe architecture, SH101 architecture, etc. As well as the ability for end users to specify and learn on their own architectures.

src/ddspdrum/module.py
```diff
@@ -365,12 +366,27 @@ def __init__(
...
 
+    @staticmethod
+    def play_vco_bank(vco_bank, pitch_envelope):
+        audio_out = 0
+        for vco in vco_bank:
+            audio_out += vco(pitch_envelope)
+        return audio_out / len(vco_bank)
+
+    @staticmethod
+    def add_noise(audio_in, ratio):
```
jorshi, line 387: Do we want to have noise as its own module?

turian: I'm fine to keep it here for now and refactor once we make a second synth besides Drum. But we can also refactor now.

*I imagined that we would have a collection of building block dsp modules, and then some preset architectures based on existing drum machines.* Agreed, different architectures but each one fixed.

maxsolomonhenry: I'm making a Mixer module that combines two VCOs and optionally adds noise. I've seen modular units like this and it seems to make sense for now.

turian: @maxsolomonhenry So I'm against this for now unless we implement it correctly.

Right now, we assume all synth modules are atomic, i.e. don't contain other synth modules.

And the synths like Drum are a list of atomic synth modules (none of them compound)

If we start adding compound synth modules, it becomes more tricky to:
* read and set parameters for a whole synth
* iterate over all modules to make sure they are wired up well (have same sample and control rates)

I argue that unless we have a careful plan for how to do these things, we should not have compound modules for now. At least until we have Synth param setting and getting over a list of atomic modules

src/ddspdrum/module.py
```diff
@@ -353,23 +396,41 @@ def __init__(
...
+                vco_2,
+                noise_module,
+                vca
+            ]
+        )
         assert sustain_duration >= 0
 
         # The convention for triggering a note event is that it has
         # the same sustain_duration for both ADSRs.
         self.pitch_envelope = pitch_adsr(sustain_duration)
```
turian, line 419: This is an old bug I just spotted that I'm fixing. We shouldn't be doing computations in __init__ because the can change when the parameters change. __init__ should only set parameters.

PR #17, opened 2021-01-08
Parameters

jorshi:
Adding named parameters with conversion to and from range [0,1] to all SynthModules as well as Synths. Can handle nested SynthModules and parameters. Controversial change of having Synth inherit from SynthModule; I found I wanted a lot of the same functionality for updating parameters within a SynthModule as well as within a Synth, and if we think we may want to support nested modules (maybe we do, maybe not) then it made sense to have these function very similarly. I realize that this may not be what we were thinking, so definitely open to conversation here.


src/ddspdrum/module.py
```diff
@@ -348,22 +430,20 @@ def __init__(
...
-        return crossfade(audio_in, noise, self.ratio)
+        return crossfade(audio_in, noise, self.parameters['ratio'].value)
 
     @staticmethod
     def noise_of_length(audio_in: np.ndarray):
         return np.random.rand(len(audio_in))
 
 
-class Synth:
+class Synth(SynthModule):
```
jorshi, line 444: This might be controversial -- I found I was duplicating a lot of the parameter updating code I wrote for the SynthModule to here, so thought that this could just be a SynthModule and use the same parameter logic as above. Easy to create named parameters that control underlying SynthModules.

src/ddspdrum/module.py
```diff
@@ -87,21 +178,11 @@ def __init__(
...
-        assert d >= 0
-        self.d = d
-
-        assert r >= 0
-        self.r = r
+        self.add_parameter("attack", a, 0, 20, scale=0.5)
+        self.add_parameter("decay", d, 0, 20, scale=0.5)
+        self.add_parameter("sustain", s, 0, 1)
+        self.add_parameter("release", r, 0, 20, scale=0.5)
+        self.add_parameter("alpha", alpha, 0, 10)
```
jorshi, line 185: Each SynthModule defines named parameters with an initial value, minimum, maximum, and scale. Scale helps in the conversion to and from 0 to 1 range

src/ddspdrum/module.py
```diff
@@ -145,24 +226,29 @@ def _ramp(self, duration: float):
...
+        return np.full(round(int(duration * CONTROL_RATE)),
+                       fill_value=self.parameters['sustain'].value)
 
     @property
     def release(self):
         # `r`-length reverse ramp, scaled and shifted to descend from `s` to 0.
-        return self._ramp(self.r)[::-1] * self.s
+        sustain = self.parameters['sustain'].value
+        release = self.parameters['release'].value
+        return self._ramp(release)[::-1] * sustain
```
jorshi, line 251: One downside is these parameters take up more room ... maybe there is a way around this?

src/ddspdrum/module.py
```diff
@@ -411,17 +500,50 @@ def __init__(
         self.noise_module = noise_module
         self.vca = vca
 
+        # Setup parameters for this drum synth
+        self.add_parameter("sustain_duration", sustain_duration, 0, 60, scale=0.5)
+        self.add_parameter("vco_1_ratio", vco_1_ratio, 0, 1)
```
maxsolomonhenry, line 505: I know it's confusing, sorry. I'd say each Drum has two "ratios":

(1) noise_ratio --> amount of noise.
(2) vco_ratio --> mix between vco 0/1. this is equivalent to a mix knob.

It might be confusing to call it "vco_1_ratio" when really it affects the level of both oscillators.

examples/examples.py
```diff
@@ -142,8 +141,35 @@ def stft_plot(signal, sample_rate=SAMPLE_RATE):
 drum_out = my_drum()
 stft_plot(drum_out)
 ipd.Audio(drum_out, rate=vca.sample_rate)
+
+# ### Parameters
+# All synth modules and synth classes have named parameters which can be quered and updated. Let's look at the parameters for the Drum we just created. Each of these parameters shows the current value, minimum, maximum, and scale. The min and max refer to the smallest and largest values that parameter can take on. The scale value controls conversion between a range of 0 and 1. Let's look at that more below.
+
+my_drum.list_parameters()
+
```
maxsolomonhenry, line 149: Nice.

src/ddspdrum/module.py
```diff
@@ -58,6 +62,94 @@ def control_to_sample_rate(self, control: np.array) -> np.array:
...
+        """
+        Add a new parameter to this module.
+
+        Parameters
+        ----------
+        parameter_id (str)  :   id to save this parameter as
+        value (float)       :   initial value for this parameter
+        minimum (float)     :   minimum value that this parameter can take
+        maximum (float)     :   maximum value that this parameter can take
+        scale (float)       :   scaling when converting between [0,1] range
```
maxsolomonhenry, line 76: Not sure I completely understand this variable. I'll read on in the code, but I suspect this comment could be a little clearer.

src/ddspdrum/parameter.py
```diff
@@ -0,0 +1,85 @@
...
+    Parameter class is a structure for keeping track of parameters
+    that have a specific range. Also handles functionality for converting
+    to and from a range between 0 and 1.
+
+    Parameters
+    ----------
+    value (float)   :   initial value of this parameter
+    minimum (float) :   minimum value that this parameter can take on
+    maximum (float) :   maximum value that this parameter can take on
+    scale   (float) :   scaling to apply when converting to and from a value with
```
maxsolomonhenry, line 19: I see. Maybe a more descriptive variable name? `curve` -- and then I would follow suit changing `alpha`

src/ddspdrum/parameter.py
```diff
@@ -0,0 +1,85 @@
...
+
+    def set_value_0to1(self, new_value):
+        """
+        Set value of this parameter using a normalized value in the range [0,1]
+
+        Parameters
+        ----------
+        new_value (float)   :   value to update parameter with, in the range [0,1]
+        """
+        new_value = np.clip(new_value, 0, 1)
```
maxsolomonhenry, line 70: @turian are we going to have gradient issues with clipping here?

src/ddspdrum/module.py
```diff
@@ -302,19 +386,18 @@ def __init__(
...
         square = np.tanh(np.pi * self.k * np.sin(argument) / 2)
-        return (1 - self.shape / 2) * square * (1 + self.shape * np.cos(argument))
+        shape = self.parameters['shape'].value
+        return (1 - shape / 2) * square * (1 + shape * np.cos(argument))
 
     @property
     def k(self):
         # What does k mean here? Can we give it a better name?
-        f0 = midi_to_hz(self.midi_f0 + self.mod_depth)
+        pitch = self.parameters['pitch'].value + self.parameters['mod_depth'].value
```
maxsolomonhenry, line 399: "Pitch" might be a little misleading here.

`k` is used to determine how many partials to calculate, and therefore is dependant on f0 and Nyquist. Higher f0, less partials.

`f0` is meant to calculate number-of-partials for a worst-case scenario. I.e., the pitch at the highest point in its modulation.

src/ddspdrum/module.py
maxsolomonhenry: Big fan of this. Nice work/

src/ddspdrum/parameter.py
turian: I don't completely understand this and need your guys advice: Do we really need the full range of curve values? Maybe we just need strings: "linear", "log", and "exp"?

Agree that the name is confusing

src/ddspdrum/parameter.py
turian: Yes, in general my preference would be that this module is not responsible for putting things in the correct range, and instead does an assert

src/ddspdrum/parameter.py
```diff
@@ -0,0 +1,85 @@
...
+            value: float,
+            minimum: float,
+            maximum: float,
+            scale: float = 1,
+            name: str = ""
+    ):
+        self.minimum = minimum
+        self.maximum = maximum
+        self.scale = scale
+        self.value = np.clip(value, minimum, maximum)
```
turian, line 39: Can we just change this to asserts?

src/ddspdrum/parameter.py
```diff
@@ -0,0 +1,85 @@
...
+        self.name = name
+
+    def __str__(self):
+        name = "{}, ".format(self.name) if self.name else ""
+        return "Parameter: {}Value: {}, Min: {}, Max: {}, Scale: {}".format(
```
turian, line 44: comma missing

src/ddspdrum/parameter.py
turian: Here and for all other clips

src/ddspdrum/module.py
```diff
@@ -411,17 +500,50 @@ def __init__(
         self.noise_module = noise_module
         self.vca = vca
 
+        # Setup parameters for this drum synth
+        self.add_parameter("sustain_duration", sustain_duration, 0, 60, scale=0.5)
+        self.add_parameter("vco_1_ratio", vco_1_ratio, 0, 1)
+
+        # Pitch Envelope
```
turian, line 507: I'm wondering if we can get rid of this boilerplate using the following approach:
* Each module has a string name, that we define when we create it
* SynthModule double checks it doesn't contain modules with duplicated name
* To address a parameter, you use a tuple: ("module_name", "parameter_name")

src/ddspdrum/module.py
turian: Agreed that I find this a bit confusing too

src/ddspdrum/module.py
```diff
@@ -58,6 +62,94 @@ def control_to_sample_rate(self, control: np.array) -> np.array:
...
+        parameter_id (str)  : Id of the parameter to update
+        value (float)       : Value to update parameter with
+        """
+        self.parameters[parameter_id].set_value_0to1(value)
+
+    def list_parameters(self):
+        """
+        Return a dictionary of the parameters as strings
+        """
+        return {key: str(self.parameters[key]) for key in self.parameters}
```
turian, line 151: maybe return it as a dict so it can be manipulated by API?

src/ddspdrum/module.py
```diff
@@ -58,6 +62,94 @@ def control_to_sample_rate(self, control: np.array) -> np.array:
...
+        maximum (float)     :   maximum value that this parameter can take
+        scale (float)       :   scaling when converting between [0,1] range
+        """
+        if parameter_id in self.parameters:
+            raise ValueError("parameter_id: {} already used".format(parameter_id))
+
+        self.parameters[parameter_id] = Parameter(value, minimum, maximum,
+                                                  scale, parameter_id)
+
+    def connect_parameter(self, parameter_id: str, module: SynthModule,
```
turian, line 84: I think we can get rid of `connect_parameter` given the strategy I'm describing below.

src/ddspdrum/module.py
```diff
@@ -58,6 +62,94 @@ def control_to_sample_rate(self, control: np.array) -> np.array:
...
+        parameter_id (str)  :   id to save this parameter as
+        value (float)       :   initial value for this parameter
+        minimum (float)     :   minimum value that this parameter can take
+        maximum (float)     :   maximum value that this parameter can take
+        scale (float)       :   scaling when converting between [0,1] range
+        """
+        if parameter_id in self.parameters:
+            raise ValueError("parameter_id: {} already used".format(parameter_id))
+
+        self.parameters[parameter_id] = Parameter(value, minimum, maximum,
```
turian, line 81: Perhaps we could simplify and in each `SynthModule.__init__` it calls ` Parameter(` multiple times instead of `add_parameter`? Then we also don't have to duplicate the documentation and maintain it twice.

The code to sanity-check we don't duplicate parameters (and that params are in the right range, etc.), I think we should maybe write generically as `SynthModule.sanity_check` and `SynthModule.__call__` will always run sanity-check before call the instance `ActualModule.__call_implementation__` or something like that.

src/ddspdrum/module.py
turian: I also find this pattern hard to grok, so am definitely looking for a way to avoid it :)

src/ddspdrum/module.py
turian: We could use:
```
def p(self, name): return self.parameters[name].value
```
which I think is an okay tradeoff given how common this idiom is in our code. Then you just do `self.p('sustain')`.

If we want to get rid of the string quotes, well we could use python black magic fuckery.

This won't work: at the end of __init__ there could be a generic SynthModule method that is called that does:

```
for p in self.parameters:
    self.__dict__[p] = self.parameters[p].value
```

the reason that won't work is because it does assign-by-value not assign-by-reference.

Instead, there might be a way to create @get and @set properties using lambda for every parameter. This might be too tricky, and perhaps just a `self.p('name')` is best

src/ddspdrum/module.py
turian: So this makes me slightly nervous, but I can live with it.

If the only reason for this is `sustain_duration` and `vco_1_ratio`, another approach would be:
* `sustain_duration` is a considered a meta-parameter, not a parameter. This is my putting on my ML hat. You would fix the sustain_duration when doing traditional ML. (However, there are more radical things where you might want to compare audio with different lengths. However, that might happen outside this anyway, and for mini-batching you'd want a fixed sustain_duration so each audio is the same length.)
* `vco_1_ratio` is part of a dummy Module that contains just a parameter that can be queried, but doesn't do any audio processing.

src/ddspdrum/parameter.py
maxsolomonhenry: I'm into this idea as well, strings.

src/ddspdrum/parameter.py
maxsolomonhenry: Cool, that's an easy fix.

turian: In terms of how you do ML when many of the parameters *should* be at the extrema of 0 or 1:
1) Pass the "preparameter" through a sigmoid-squashing function to turn it into a synth parameter. This is good IF you are okay with the the parameter getting close to 0 / 1 but never hitting it.
2) Pass the "preparameter" through a RELU-like squashing function to clip it, basically. The downside is you get no gradient (like dead neurons) so I think this is a no-go.
3) Add a steep penalty to the loss function if the parameter goes outside [0 - eps, 1 + eps], eps is maybe like 1e-4. Before you do a forward, the ML clips the parameters to 0, 1, *not the synth stuff*. Then it can get a gradient that is like: "Okay we want this parameter less than 0" from the synth but also know "okay but if I make this parameter further from 0 I get penalized so keep the parameter within eps of 0 and that sweet spot means the parameter is actually 0". This is the most promising approach I've tried.

src/ddspdrum/module.py
turian: Alternately, each Synth contains a ParametersModule which is just parameters.

src/ddspdrum/module.py
maxsolomonhenry: I like this idea too. I can't quite picture it -- @jorshi do you have enough of a handle to implement it?

src/ddspdrum/module.py
jorshi: Sure, so just add Parameters directly to self.parameters in `__init__`? 

Not sure if I totally understand how the sanity check function would work.

src/ddspdrum/module.py
jorshi: Idea here is to be able to get a quick look at what the parameters are, if you check out the example you get a nice print out of the parameter values / min / max in Jupyter -- maybe this would make more sense to be renamed print_parameter or something like that? Parameters can be accessed as a dict directly through the parameters member.

src/ddspdrum/module.py
jorshi: `def p(self, name): return self.parameters[name].value` -- I like this

src/ddspdrum/module.py
jorshi: Yes makes sense -- will rename to something more meaningful.

src/ddspdrum/parameter.py
jorshi: There might be some parameters where were want a more extreme curve. But I am also game to simplify things with strings -- let's go for that.

src/ddspdrum/parameter.py
jorshi: Yes definitely.

src/ddspdrum/module.py
jorshi: Sounds good

src/ddspdrum/module.py
jorshi: It wasn't just for the `sustain_duration` and `vco_1_ratio` -- also to handle the nested parameters within the SynthModules. I kind of like thinking about a Synth as just a SynthModule that contains nested SynthModules, could lead to some interesting architectures. That being said, with the comments on how to remove connect_parameters and the idea of having a ParametersModule, it would totally work to remove the SynthModule inheritance -- might be better to keep relationships simplified.

src/ddspdrum/module.py
jorshi: Sweet, ya I think I've got an idea of how to implement that.

src/ddspdrum/module.py
turian: yeah easy and concise

src/ddspdrum/parameter.py
turian: Yeah the nice thing about strings if is we want to add one more, we can easily add it to the switch-statement

src/ddspdrum/module.py
turian: Sure, maybe this should just be the `__repr__` for the class instance? The long complete string.

https://stackoverflow.com/questions/1436703/difference-between-str-and-repr

* __repr__ goal is to be unambiguous
* __str__ goal is to be readable

src/ddspdrum/module.py
turian: @jorshi don't worry about the sanity check, I could implement it in another PR.

Yeah the simple idea is that, by convention, all __init__ directly create `Parameters`. This traces a nice clean line to a PyTorch port where every `nn.Module` will init its torch parameters in `__init__`. If you're not sure about this I can tackle it.

src/ddspdrum/module.py
turian: I think let's keep it flat for now, let's get a few more things nailed down, and then let's make some cool nested architectures soon

src/ddspdrum/module.py
maxsolomonhenry: +1

src/ddspdrum/parameter.py
maxsolomonhenry: I think @jorshi's point is that you might want to continuously adjust the severity of the curve. I get that. But I don't know how much of a perceptual difference it will make, and it might save us some headache just to constrict the problem at some point.

src/ddspdrum/module.py
turian: Implemented here: https://github.com/turian/ddsp-drum/pull/21

src/ddspdrum/module.py
turian: I added this https://github.com/turian/ddsp-drum/pull/20 but see discussion we were having about how we can't make Synth not a SynthModule and remove `connect_parameter` without making Synth have its own business logic for compound parameter addressing.

src/ddspdrum/module.py
turian: Let's just figure out whether we want to use tuples or connect_parameter and strings for compound parameter addressing, before we tackle this.

src/ddspdrum/parameter.py
turian: @maxsolomonhenry You don't adjust `scale` during training, I don't think that's @jorshi 's proposal. These are decided offline by the people implementing the modules (i.e. us) during design time, using expert knowledge about the perceptual phenomena around each parameter. Is my current understanding of the pattern we're using.

src/ddspdrum/module.py
turian: I removed it https://github.com/turian/ddsp-drum/pull/20 that these are no longer add_parameter.

However, if we stay with the pattern of `connect_parameter` instead of `tuples`, we don't need my change per se.

src/ddspdrum/parameter.py
jorshi: Ya this wouldn't be updated during training -- but some users might want to be able to adjust the severity of the curve. I think we're fine to stick with log, exp, or linear for the time being.

src/ddspdrum/module.py
jorshi: Looks like `__repr__` is intended for debugging / logging and is supposed to show everything in class? Intention of this function is to get a quick snapshot into the parameters and their names for end-users. If I'm working with a SynthModule in a Jupyter notebook and I want to see what parameters are available and their range then I could call this function. Do you think `__repr__` is okay for that?

turian: For me __repr__ seems like the natural choice, but if you are uneasy about it or think it will involve a lot more chatty text, then we can skip it.

turian (approved):

PR #22, opened 2021-01-09
Parameter cleanup

jorshi:
Addressing a few minor changes to the parameters PR. Renaming the parameter scale to curve and fixing to linear, log, or exp. Removed clipping of value in the Parameter class. Renamed the k property in SquareSawVCO to be more descriptive.

### DIFF
--- a/src/ddspdrum/parameter.py
+++ b/src/ddspdrum/parameter.py
@@ -23,12 +23,12 @@ class Parameter:
     """
 
     def __init__(
-            self,
-            name: str,
-            value: float,
-            minimum: float,
-            maximum: float,
-            curve: str = "linear",
+        self,
+        name: str,
+        value: float,
+        minimum: float,
+        maximum: float,
+        curve: str = "linear",
     ):
         self.name = name
         self.minimum = minimum
--- a/src/ddspdrum/parameter.py
+++ b/src/ddspdrum/parameter.py
@@ -16,12 +16,9 @@ class Parameter:
     value (float)   :   initial value of this parameter
     minimum (float) :   minimum value that this parameter can take on
     maximum (float) :   maximum value that this parameter can take on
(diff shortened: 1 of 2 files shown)


src/ddspdrum/parameter.py
```diff
@@ -57,7 +65,7 @@ def set_value(self, new_value):
         ---------
         new_value (float)   :   value to update parameter with
         """
-        self.value = np.clip(new_value, self.minimum, self.maximum)
+        self.value = new_value
```
jorshi, line 68: Should we assert here as well?

src/ddspdrum/parameter.py
```diff
@@ -67,10 +75,8 @@ def set_value_0to1(self, new_value):
         ----------
         new_value (float)   :   value to update parameter with, in the range [0,1]
         """
-        new_value = np.clip(new_value, 0, 1)
-
-        if new_value != 0 and self.scale != 1:
-            new_value = np.exp2(np.log2(new_value) / self.scale)
+        if new_value != 0 and self.curve != 1:
+            new_value = np.exp2(np.log2(new_value) / self.curve)
```
jorshi, line 79: No asserting here either. I assume we want to avoid throwing an assertion error during learning in the case it is slightly above or below 0 or 1?

src/ddspdrum/parameter.py
turian: yes, everywhere there's a clip it indicates a precondition to assert, IMHO

src/ddspdrum/parameter.py
turian: Yes. The learner module should be responsible for preclipping.

See my suggestion here: https://github.com/turian/ddsp-drum/pull/17#discussion_r554015001
About how we should separate responsibilities

turian (approved):

PR #29, opened 2021-01-14
Envelope behaviour

maxsolomonhenry:
See examples for an outline of the new ADSR behaviour. Reacts basically as it would to midi input. Also, setting sustain_duration to 0 gives a one-shot behaviour, same as we've been working with so far.


examples/examples.py
```diff
@@ -38,14 +38,12 @@ def stft_plot(signal, sample_rate=SAMPLE_RATE):
 
 
 # Synthesis parameters.
-a = 0.01
-d = 0.05
+a = 0.1
+d = 0.1
```
maxsolomonhenry, line 42: Extended this to better demonstrate new env behaviour.

examples/examples.py
```diff
@@ -58,12 +56,27 @@ def stft_plot(signal, sample_rate=SAMPLE_RATE):
...
 # this notebook we will use the same envelope to modulate several different audio parameters.
+#
+# ### A note about note-on, note-off behaviour 
+#
+# By default, this envelope reacts as if it was triggered with midi, for example playing a keyboard. Each midi event has a beginning and end: note-on, when you press the key down; and note-off, when you release the key. `sustain_duration` is the amount of time that the key is depressed. During the note-on, the envelope moves through the attack and decay sections of the envelope. This leads to musically-intuitive, but programatically-counterintuitive behaviour.
+#
+# Assume attack is 0.5 seconds, and decay is 0.5 seconds. If a note is held for 0.75 seconds, the envelope won't traverse through the entire attack-and-decay phase (specifically, it will execute the entire attack, and 0.25 seconds of the decay).
+#
+# If this is confusing, don't worry about it. ADSR's do a lot of work behind the scenes to make the playing experience feel natural. Alternately, you may specify one-shot mode (see below), which is more typical of drum machines.
```
maxsolomonhenry, line 67: ^ this is the basic spiel for the new behaviour. should be clear enough.

src/ddspdrum/module.py
```diff
@@ -1,7 +1,7 @@
 """
 Synth modules.
 
-    TODO    :   - ADSR needs fixing. sustain duration should actually decay.
+    TODO    :   - sustain_duration change name?
```
maxsolomonhenry, line 4: I'm open to suggestions here...

src/ddspdrum/module.py
```diff
@@ -166,20 +165,47 @@ def __init__(
...
-                self.note_off,
+
+        # If sustain is "0" go to one-shot mode (moves through ADR sections).
+        if sustain_duration == 0:
+            num_samples = round(
+                int(self.p("attack") + self.p("decay") * SAMPLE_RATE)
+            )
+        else:
+            num_samples = round(
+                int(sustain_duration * SAMPLE_RATE)
```
maxsolomonhenry, line 202: This feels a bit clunky, but if it's readable I'm happy to leave it.

src/ddspdrum/module.py
```diff
@@ -166,20 +165,47 @@ def __init__(
...
             )
-        )
+
+        ADS = self.note_on(num_samples)
+        R = self.note_off(ADS[-1])
+
```
maxsolomonhenry, line 207: This is the star of the show. Under the new behaviour:

Holding a note passes through A and D stages. If there's time left over, it sustains at the `s` value. This is just like a midi instrument.

Release acts as before, except it decays from the last value of ADS (which may not necessarily be `s`, depending on how long the note was held for).

src/ddspdrum/module.py
```diff
@@ -166,20 +165,47 @@ def __init__(
...
         """
 
         assert sustain_duration >= 0
-        return np.concatenate(
-            (
-                self.note_on,
-                self.sustain(sustain_duration),
-                self.note_off,
+
+        # If sustain is "0" go to one-shot mode (moves through ADR sections).
```
turian, line 195: I'm slightly concerned about the non-smooth transition between one-shot and sustained behavior. Perhaps this is between modulated by a "one_shot: Bool" param?

src/ddspdrum/module.py
turian: Perhaps this comment should be a comment in code?

examples/examples.py
```diff
@@ -58,12 +56,27 @@ def stft_plot(signal, sample_rate=SAMPLE_RATE):
...
+#
+# If this is confusing, don't worry about it. ADSR's do a lot of work behind the scenes to make the playing experience feel natural. Alternately, you may specify one-shot mode (see below), which is more typical of drum machines. 
 
 # Envelope test
 adsr = ADSR(a, d, s, r, alpha)
 envelope = adsr(sustain_duration)
 time_plot(envelope, adsr.sample_rate)
 
+# ### One-Shot Mode
+#
```
turian, line 74: The fact that you call this a mode suggests it should be a separate param, and not overload the meaning of sustain_duration with a non-smooth (non-differentiable) transition of behavior.

src/ddspdrum/module.py
```diff
@@ -166,20 +165,47 @@ def __init__(
...
+        key. `sustain_duration` is the amount of time that the key is depressed.
+
+        During the note-on, the envelope moves through the attack and decay
+        sections of the envelope. This leads to musically-intuitive, but
+        programatically-counterintuitive behaviour:
+
+        Assume attack is 0.5 seconds, and decay is 0.5 seconds. If a note is
+        held for 0.75 seconds, the envelope won't traverse through the entire
+        attack-and-decay (specifically, it will execute the entire attack, and
+        0.25 seconds of the decay).
```
turian, line 182: Understood. I'm slightly leaning towards, based upon discussion @jorshi was in, that we just assert if the params are not programmatically sensible. i.e. assert attack + decay <= sustain_duration

src/ddspdrum/module.py
jorshi: Perhaps `note_on_duration`?

src/ddspdrum/module.py
jorshi: Agreed -- could be a util function, that might be useful later

src/ddspdrum/module.py
maxsolomonhenry: Yes, I initially was worried about this. Then I remembered that we're considering `sustain_duration` a hyper-parameter that we won't backprop through. So I'm less concerned.

Does that address your concern?

examples/examples.py
maxsolomonhenry: ^ see above. Again, note-on time is a hyperparameter so I'm not worried about it. Plus even if it were a seperate param it would still be a discrete one, which is not learnable.

src/ddspdrum/module.py
maxsolomonhenry: It's often sensible to hold a note-on for less than A + D times. Happens all the time when you're playing an instrument with a fixed envelope (ie. 99.9999% of the case)

src/ddspdrum/module.py
maxsolomonhenry: Done.

src/ddspdrum/module.py
maxsolomonhenry: Got it.

src/ddspdrum/module.py
maxsolomonhenry: Ok

examples/examples.py
```diff
@@ -1,8 +1,7 @@
 # # ddsp-drum examples
 #
-# We walk through basic functionality of `ddsp-drum` in this Jupyter notebook. Just note that all ipd.Audio play widgets normalize the audio.
-#
-# In this example, we'll build a very basic kick sound.
```
maxsolomonhenry, line 5: No longer really doing this. Could return to this idea later.

src/ddspdrum/module.py
```diff
@@ -45,6 +44,9 @@ def __repr__(self):
             self.__class__, repr(self.sample_rate), repr(self.parameters)
         )
 
+    def seconds_to_samples(self, seconds: float) -> int:
+        return round(int(seconds * self.sample_rate))
+
```
maxsolomonhenry, line 49: Controversial, again. This looks like the best place for this function.

It didn't make sense to put it in utils, it's tied to a sample rate, which is always carried in the object.

Otherwise you get a call like `seconds_to_samples(seconds_val, self.sample_rate)` which doesn't save you much space.

src/ddspdrum/module.py
```diff
@@ -165,21 +167,46 @@ def __init__(
...
-            )
-        )
+        assert note_on_duration >= 0
+
+        # If sustain is "0" go to one-shot mode (moves through ADR sections).
+        if note_on_duration == 0:
+            note_on_duration = self.p("attack") + self.p("decay")
+
+        num_samples = self.seconds_to_samples(note_on_duration)
+
```
maxsolomonhenry, line 204: 200 - 203 feels a lot cleaner now.

src/ddspdrum/module.py
turian: That's a good point. Could you maybe just summarize my concern and your rebuttal as a comment in code and we can move on?

turian: @maxsolomonhenry actually see comment below

examples/examples.py
turian: @maxsolomonhenry main question tho: If I gave you sound and asked you to predict what patch generated it, we'd probably want to include one-shot vs not-one-shot in the parameter space.

If that were the case then learning a boolean parameter would be appropriate. you are right that learning a hard discrete parameter is difficult through gradient descent. but if we relaxed it and turned it into a [0,1] mix parameter between one-shot and not-one-shot mode, then we could. this is one of the cases where the synth becomes a bit more patchable than most humans would want but perhaps a human would want

src/ddspdrum/module.py
turian: okay, sounds good

src/ddspdrum/module.py
turian: I'm fine with this

maxsolomonhenry:
Not sure what this issue is with coverage. But it seems minor. Pulling 'er in so we can start pytorching it.

PR #30, opened 2021-01-21
added convenience import

maxsolomonhenry:
You can take this or leave it. I've been exploring a bit the python package structure.

in essence, this allows for the example.py import:

```from ddspdrum.module import ADSR  # etc...```

to become:

```from ddrpdrum import ADSR # etc...```

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -22,7 +22,7 @@ jobs:
           use-isort: true
           extra-pylint-options: ""
           extra-pycodestyle-options: ""
-          extra-flake8-options: "--max-line-length=88 --extend-ignore=E203"
+          extra-flake8-options: "--max-line-length=88 --extend-ignore=E203 --per-file-ignores=__init__.py:F401"
           extra-black-options: ""
           extra-mypy-options: ""
           extra-isort-options: ""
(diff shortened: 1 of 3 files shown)


jorshi:
I'm happy with this change -- I think it makes sense given the scope of this project. The modules are the main components so it makes sense to have them accessible at the root package level. Just fix up the linter error and good from my perspective!

maxsolomonhenry:
Awesome.

You'll have to forgive my n00biness... I'm not sure how to read these bot readouts. Is there a doc you could point me to?

Max
________________________________

maxsolomonhenry:
Ah, I see. It's saying 

`src/ddspdrum/__init__.py:1:1: F401 'ddspdrum.module.ADSR' imported but unused` etc... in  `__init__.py` but there's nothing to use in those modules for in the init file... So I'm not sure what to do here.

turian:
@jorshi is the pythonic way that __init__.py just imports the submodules, rather than the things in it?

If not, @maxsolomonhenry I think the answer is just to add a pylint disable. But let's see what @jorshi says

jorshi:
I've seen arguments for both having empty `__init__` files as well as importing specific things from with submodules. Argument for having empty `__init__`s seems to be that they can become complex as the library grows. I think it makes sense here to import the main items (SynthModules / TorchModules) in the `__init__` and keep the other stuff under their submodule namespace.  i've seen that pattern in other libraries like librosa. So in that case we can add the pylint disable and move forward with this.

PR #31, opened 2021-01-21
Tests

jorshi:
Starting unit tests with PyTest

### DIFF
--- a/.github/workflows/codecov.yml
+++ b/.github/workflows/codecov.yml
@@ -49,13 +49,13 @@ jobs:
         pytest --cov-report=xml --cov=./src tests/
     - name: Upload pytest coverage to Codecov
       uses: codecov/codecov-action@v1.2.0
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
-          files: ./coverage.xml
-          #        directory: ./coverage/reports/
-          flags: unittests
-          env_vars: OS,PYTHON
-          name: codecov-umbrella
-          fail_ci_if_error: true
-          #        path_to_write_report: ./coverage/codecov_report.txt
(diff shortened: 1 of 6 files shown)


src/ddspdrum/module.py
```diff
@@ -45,7 +45,7 @@ def __repr__(self):
         )
 
     def seconds_to_samples(self, seconds: float) -> int:
-        return round(int(seconds * self.sample_rate))
+        return int(round(seconds * self.sample_rate))
```
jorshi, line 48: I think we want to round before converting to int here

.github/workflows/codecov.yml
```diff
@@ -42,3 +44,18 @@ jobs:
         fail_ci_if_error: true
 #        path_to_write_report: ./coverage/codecov_report.txt
         verbose: true
+    - name: Generate coverage report with tests
```
jorshi, line 47: Adding pytest to codecov report

.github/workflows/pytest.yml
```diff
@@ -0,0 +1,31 @@
+name: Tests with Pytest
```
jorshi, line 1: maybe redundant adding a pytest specific workflow? But we can add tests for different OSes and python versions in addition to the one run in the code coverage

tests/test_module.py
```diff
@@ -0,0 +1,124 @@
+"""
```
jorshi, line 1: pytest looks for modules prefixed with test_ and then runs methods prefixed with test_ as unit tests. Grouping each test within a class that relates to the class that it is testing.

turian (approved):

PR #33, opened 2021-01-23
Rename parameter to modparameter

turian:
Okay so I'm not sure I love this. However, if TorchSynthModule will inherit from SynthModule and torch.nn.Module, I don't want to break torch. And torch has a special `parameters` class member for each `nn.module`.

Let's also see what happens when we have TorchSynthModules with Torch parameters. That might give us an easy way to address submodule parameters.

The nice thing is this is easy to search and replace, particularly using pycharm, if we want to change this to something nicer later.

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -190,7 +190,7 @@ print(my_drum)
 
 # Setting a parameter with a range of [0,1]
 
-my_drum.set_parameter_0to1("pitch_attack", 0.25)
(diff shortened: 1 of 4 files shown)


maxsolomonhenry:
I don't love it. `Parameter` is clear -- are we really at risk of collisions here? SUrely lots of modules have local so-named parameters?

If we do have to change it I'd go for somthing more descriptive. "synthparameter" or something

PR #36, opened 2021-01-23
npyforward instead of __call__

turian:
All __call__ commands are now npyfoward. I'm not 100% sure but I believe will need this to wrap, as I'm writing the TorchSynthModule.forward.

The nice thing is this is easy to search and replace, particularly using pycharm, if we want to change this to something nicer later.


maxsolomonhenry:
I think I understand what you're doing here. Is the idea that each module will have a vestigial  `npyforward` in numpy, and a `__call__` in pythorch?

PR #38, opened 2021-01-23
Convert params to floats

turian:
Builds on #33 

We don't want things that are floats actually to be ints. This can trip up pytorch.


maxsolomonhenry (commented):
Yeah again, fine with this contingent on naming decisions etc form other PR's.

PR #39, opened 2021-01-23
Enable mypy for type checking

turian:


turian:
I think this will be a headache to get working and won't help much, once everything is in torch :\

PR #41, opened 2021-01-23
torchutil.py

turian:
* I port util.py to torchutil.py
* util.py is now numpyutil.py
* Unit tests do fuzz testing to make sure that numpy and torch utils are relatively within 1e-4 of each other.

**Ignore code coverage for now.** Code coverage might drop for a little, since we have unit-tests but not integration tests for the torch stuff

### DIFF
--- a/src/ddspdrum/torchutil.py
+++ b/src/ddspdrum/torchutil.py
@@ -1,0 +1,76 @@
+"""
+Utility functions for torch DSP related things
+
+TODO: After everything is torch'ified, remove numpyutil.py version
+and rename this to util.py.
+
+TODO: These should operate on vectors, many of these assume scalar Tensors.
+"""
+
+import torch
+import torch.tensor as T
+
+from ddspdrum.defaults import EPSILON, EQ_POW
+
+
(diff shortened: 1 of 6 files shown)


maxsolomonhenry (approved):
Beaut'

src/ddspdrum/torchutil.py
```diff
@@ -0,0 +1,76 @@
...
+"""
+
+import torch
+import torch.tensor as T
+
+from ddspdrum.defaults import EPSILON, EQ_POW
+
+
+# What is amin here? And maybe we should convert it to a value in defaults?
+# What is the range of amplitude?
```
jorshi, line 17: amin is the minimum amplitude to avoid taking the log of zero -- it could certainly be a value in defaults. Expecting an amplitude in the range of 0 to 1 which maps to a dB range of -200dB to 0dB when using amin=1e-10. (Amplitude could be greater than 1 which would give a dB value greater than zero, but since we are dealing with float samples between -1 and 1 then that would not be expected)

maxsolomonhenry (approved):
Let's pull er in?

PR #42, opened 2021-01-23
Randomize modparameter

turian:
This allows you to create randomized parameters for the synth. I'd like this for unit-testing numpy vs torch.

Depends upon #33 

What's weird is that the note_on_duration is sometimes disregarded and you get like 16 second samples. Is this because of the one-hit thing @maxsolomonhenry ? That seems like a problem to me. You should be able to hillclimb the synth by setting random params to try to get a similar audio of the same duration :\

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -190,7 +190,7 @@ print(my_drum)
 
 # Setting a parameter with a range of [0,1]
 
-my_drum.set_parameter_0to1("pitch_attack", 0.25)
+my_drum.set_modparameter_0to1("pitch_attack", 0.25)
 print(my_drum.parameters['pitch_attack'])
 
 drum_out = my_drum()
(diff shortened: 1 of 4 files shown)


maxsolomonhenry:
As per [this issue](https://github.com/turian/ddsp-drum/issues/27#issue-786093911) we need to reconsider our parameter min-maxs. A 20 second maximum attack will get you into this mess.

turian:
@maxsolomonhenry The issue is orthogonal to the one you're talking about. It's more than if you have a SynthModule (or Drum) and call "randomize_parameters", you don't expect the sound length to change. One you have a synth set up and start flipping around the parameters to find a particular sound, you don't want the sound length to vary.

maxsolomonhenry:
Again we're not hillclimbing note_on_duration. That's a hyperparameter.

If you're randomizing, the audio could have a max time value, based on the Parameter max values.

jorshi:
Randomizing the release time parameter would lead to samples with different lengths, right? Max length of the sample would be note_on_duration + maximum release time. One thing we could do to fix the length of the sound is to introduce another hyperparameter, something like `render_length` -- this could set the length of the output. For example, if the note_on_duration was 1s and that render_length was 2s, then the release time portion would just be cut off at 1s regardless of the actual release time.

turian:
@jorshi yes I like the idea of a render_length hyperparameter.

PR #37, opened 2021-01-23
Refactors before torch-ifying

turian:
This combines #33 #35 #36 #38 #42

PR #44, opened 2021-01-23
Torch ADSR + SineVCO

turian:
This PR is diff'ed against #37 for better understanding.

In this PR, I am converting ADSR and SineVCO to torch. I make sure that torch and numpy modules give the same values on forward. Gradients are computed in example.py but not in unit tests. Why are we getting nan gradient for alpha?

There's a lot to hate in this port.

~~In general, before we address the nitty-gritty (below), I think the biggest issue to consider is that I don't know if standard ADSR will be differentiable on the adsr parameters, since it involves a lot of discontinuities and padding. I think we have two options~~
~~1) don't make anything differentiable and just focus on rendering speed. this is quite lame, of course~~
~~2) work on the subproblem of just creating a differentiable ADSR. ignore our complicated abstractions for now, and just focus on differentiable ADSR, maybe to minimize a simple l2 distance. maybe we have to use splines?~~

Update: It turns out I can differentiate through ADSR parameters? Except alpha which is nan. Why? Check out `example.py`

Here's some TODO:
* examples.py should have a torch section doing the exact thing as above, but with the torch versions.
* Internally, we want to be storing all nn.Parameter in the 0/1 range, not the human readable range. Since that is what will be used for backprop.
* The modparameter abstraction needs to be cleaned up, since each nn.Parameter stores its own value. I think the cleanest thing is a TorchModParameter which inherits from nn.Parameter but adds some helper methods around it.
* torch.linspace doesn't have an endpoint param.
* gradients should be a unit test


src/ddspdrum/torchmodule.py
```diff
@@ -0,0 +1,406 @@
...
+
+        Its curve is determined by alpha:
+
+        alpha = 1 --> linear,
+        alpha > 1 --> exponential,
+        alpha < 1 --> logarithmic.
+
+        """
+
+        # TODO: We originally used endpoint=False, which torch.linspace doesn't support :(
```
turian, line 255: @maxsolomonhenry thoughts?

turian:
One more thing we missed: We have Parameter so that we can query and control and randomize the synth settings through API. However, we don't have that for the *forward* parameters (like note_on_duration) of each synth module.

This is annoying because I'd like to just write unit tests that completely fuzz test numpy vs torch synth modules, and can sensibly automagically pick random forward variables for them. The range of each forward parameter could be specified as list of constants in SynthModule defined before `__init__`.

This is leading me to believe that Parameter (which I'm renaming to ModParameter to not clobber nn.Parameter) should not store a value, and should just be a meta-description of the parameter scale. That seems to be the pattern I want for torch modules and also for randomizing the forward parameters.

src/ddspdrum/torchmodparameter.py
```diff
@@ -0,0 +1,97 @@
+"""
+Parameter Class
```
jorshi, line 2: What is this module for? Looks like TorchSynthModule is still using the other ModParameters?

src/ddspdrum/torchmodule.py
```diff
@@ -0,0 +1,410 @@
...
+            else:
+                npyinput.append(i)
+        return self.forward(*npyinput).numpy()
+
+    # The following is cheesy AF but needed because the
+    # torchparameter is the master variable.
+    # TODO: Remove .value from modparameter
+    def _update_modparameters(self) -> None:
+        for modparameter_id in self.modparameters:
+            self.modparameters[modparameter_id].set_value(
```
jorshi, line 90: This should be set_value_0to1 no? That is the range we are getting back from torch parameters correct ([0,1])?

src/ddspdrum/torchmodparameter.py
jorshi: I see now that the plan is for this to inherit from nn.Parameter

jorshi:
Good stuff! The parameter stuff feels like it is getting a bit messy for sure. I realize this is just a step towards the realized version, which is great. I like the idea of getting rid of ModParameter holding onto any value and creating a subclass of nn.Parameter to handle that. We could rename ModParameter to ParameterRange or something similar and have it be responsible for converting between the human range and the internal 0-1 range? This could either live within the TorchModParameter (maybe we can lose the Mod part of the name now?) -- or just be part of a TorchSynthModule and be associated with a TorchParameter?

Also, I ran examples on my machine and am getting real values for the alpha gradient.

src/ddspdrum/torchmodparameter.py
turian: you're right tho we can remove it

src/ddspdrum/torchmodule.py
turian: @jorshi no, right now we are wrapping the existing behavior so torchparameters are in the original range. I do think we should change it to 0, 1 internally tho

turian:
@jorshi yeah the proposed rewrite ideas are good. I think parameter was a good step forward for us, we weren't 100% sure and needed to figure out the architecture as we went.

Can you paste the gradients you're getting?

src/ddspdrum/torchmodule.py
maxsolomonhenry: Yeah just looking into this. We could make a wrapper for torch linspace that does the extra calculations to ignore the endpoint. Or we could just plug our ears and hope this doesn't make a big difference.

PR #54, opened 2021-01-24
linspace func in torchutil w/ endpoint option

maxsolomonhenry:
Pretty straightforward. Allows to keep the `np.linspace` behaviour we've been using and keeps the math nice. The only issue is the choice of function name. Currently it's `linspace` straight up.

`linspace_with_option_endpoint` is clear, but long. Another choice is `my_linspace` or something like that.

### DIFF
--- a/src/ddspdrum/torchmodule.py
+++ b/src/ddspdrum/torchmodule.py
@@ -12,7 +12,7 @@ import torch.tensor as T
 
 from ddspdrum.defaults import SAMPLE_RATE
 from ddspdrum.modparameter import ModParameter
-from ddspdrum.torchutil import midi_to_hz, reverse_signal
+from ddspdrum.torchutil import midi_to_hz, reverse_signal, linspace
 
 torch.pi = torch.acos(torch.zeros(1)).item() * 2  # which is 3.1415927410125732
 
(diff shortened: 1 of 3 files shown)


turian:
Could you use linspace where I have the comment we want linspace = False?

Also, could you write a unit test in test_torchutil to fuzz test it's good?

maxsolomonhenry:
I can but we have to merge this first -- there's no `tochmodule.py` in this branch

turian:
@maxsolomonhenry I added an optional atol parameter, since relative tolerance is too sensitive for small values

maxsolomonhenry:
Great, thanks. Looks like everything else is just issues with the main repo, which we'll have to clean up eventually in the torchmodule.py etc. So good to pull?

src/ddspdrum/torchmodule.py
```diff
@@ -12,7 +12,7 @@
 
 from ddspdrum.defaults import SAMPLE_RATE
 from ddspdrum.modparameter import ModParameter
-from ddspdrum.torchutil import midi_to_hz, reverse_signal
+from ddspdrum.torchutil import midi_to_hz, reverse_signal, linspace
```
jorshi, line 15: Linter is complaining about the ordering of imports, wants them to be alphabetical

maxsolomonhenry: There's something a little S&M about this linting workflow.

PR #56, opened 2021-01-25
fixed buffer_size for renders

maxsolomonhenry:
Introducing `buffer_size`, which is also a global default `BUFFER_SIZE`.

Now `SynthModule` and `TorchSynthModule` have this property, and a method `to_buffer_size()`. I've placed this at the return for every foward/call:

```
def forward # or npyforward...
  # ...
  out_ = # previous output
  return self.to_buffer_size(out_)
```

There might be a cleaner way of doing this, like making a "pre-forward" method, and then having "forward" always be:
```return self.to_buffer_size(self.pre_forward)```
or some such thing, but that seems confusing and over-engineered.

I fixed up all of `module.py`, and adapted all of `torchmodule.py` that currently exists.

Lmk.

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -390,10 +390,6 @@ kick = svf.npyforward(signal, cutoff_mod=cutoff_mod, cutoff_mod_amount=150)
 plt.plot(kick)
 ipd.Audio(kick, rate=sample_rate)
 
-
-
-
-
 # ## Torch examples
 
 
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by turian** [comment]
I do think it might be less error prone if synthmodule and torchsynthmodule applies the command automatically on _forward in forward (or npyforward)

**Comment by maxsolomonhenry** [comment]
I'm into it. What would that look like, with the PyTorch api?


________________________________

**Comment by jorshi** [comment]
I think this is maybe what Joseph was getting at, but we could declare an abstract method `_forward` in `SynthModule` and `TorchSynthModule` that is implemented in all inheriting classes. Then the `forward` method would look something like:

```
def forward(args)
    out_ = self._forward(args)
    return self.to_buffer_size(out_)
```

Is that sort of what you had in mind @turian?

turian (approved):

PR #59, opened 2021-01-27
TorchParameter

jorshi:
Parameters for TorchSynthModules that have an internal range from 0 to 1 and can hold a ParameterRange object to convert to and from a user specified range

### DIFF
--- a/src/ddspdrum/parameter.py
+++ b/src/ddspdrum/parameter.py
@@ -57,8 +57,10 @@ class ParameterRange:
         value (float)   : value within [0,1] range to convert to range defined by
             minimum and maximum
         """
-        assert 0 <= value <= 1
-        if value != 0 and self.curve != 1:
+        assert torch.all(0.0 <= value)
+        assert torch.all(value <= 1.0)
+
+        if self.curve != 1:
             value = torch.exp2(torch.log2(value) / self.curve)
 
         return self.minimum + (self.maximum - self.minimum) * value
@@ -71,7 +73,9 @@ class ParameterRange:
         ----------
         value (float)   : value within the range defined by minimum and maximum
         """
-        assert self.minimum <= value <= self.maximum
+        assert torch.all(self.minimum <= value)
+        assert torch.all(value <= self.maximum)
+
         normalized = (value - self.minimum) / (self.maximum - self.minimum)
         if self.curve != 1:
             normalized = torch.pow(normalized, self.curve)
@@ -130,5 +134,4 @@ class TorchParameter(nn.Parameter):
         if self.parameter_range is not None:
             self.data = self.parameter_range.to_0to1(new_value)
         else:
-            assert 0 <= new_value <= 1
-            self.data = new_value
+            raise RuntimeError("A range was never set for this parameter")
--- a/src/ddspdrum/parameter.py
+++ b/src/ddspdrum/parameter.py
@@ -5,7 +5,6 @@ Parameters for DDSP Modules
 import torch
 import torch.tensor as T
 import torch.nn as nn
-import numpy as np
 
 
 class ParameterRange:
@@ -49,7 +48,7 @@ class ParameterRange:
             self.minimum, self.maximum, self.curve_type
         )
 
-    def from_0to1(self, value) -> float:
+    def from_0to1(self, value: T) -> T:
         """
         Set value of this parameter using a normalized value in the range [0,1]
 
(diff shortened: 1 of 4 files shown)


src/ddspdrum/parameter.py
```diff
@@ -0,0 +1,159 @@
...
+        """
+        Set value of this parameter using a normalized value in the range [0,1]
+
+        Parameters
+        ----------
+        value (float)   : value within [0,1] range to convert to range defined by
+            minimum and maximum
+        """
+        assert torch.all(0.0 <= value)
+        assert torch.all(value <= 1.0)
```
jorshi, line 61: @turian how should we handle values that are beyond this range? I was doing some tests with optimization and was getting errors because the parameters were going beyond this.

turian: I am not 100% sure.

As I wrote up earlier, I think what we want to do during training is allow to be epsilon (1e-1?) outside of [0, 1] without any penalty. And once it's smaller than -eps or greather than 1+eps, there is a strong penalty like 10 * amount outside range.

It's not entirely clear to me yet whether we should allow things outside the range here or not, or how to implement it. We can chat a little bit about it but this is fine for now and we can fix it later if we decide something else

src/ddspdrum/parameter.py
```diff
@@ -0,0 +1,159 @@
...
+        else:
+            curve_types = ["linear", "log", "exp"]
+            raise ValueError("Curve must be one of {}".format(", ".join(curve_types)))
+
+    def __repr__(self):
+        return "ParameterRange(min={}, max={}, curve={})".format(
+            self.minimum, self.maximum, self.curve_type
+        )
+
+    def from_0to1(self, value: T) -> T:
```
turian, line 51: Should we call this variable `normalized`? For consistency with the below?

src/ddspdrum/parameter.py
```diff
@@ -0,0 +1,159 @@
...
+        assert torch.all(self.minimum <= value)
+        assert torch.all(value <= self.maximum)
+
+        normalized = (value - self.minimum) / (self.maximum - self.minimum)
+        if self.curve != 1:
+            normalized = torch.pow(normalized, self.curve)
+
+        return normalized
+
+
```
turian, line 85: Very cool

src/ddspdrum/parameter.py
```diff
@@ -0,0 +1,159 @@
...
+        self.parameter_name = parameter_name
+
+        return self
+
+    def __repr__(self):
+        return "TorchParameter(name={}, value={})".format(
+            self.parameter_name, self.item()
+        )
+
+    def get_in_range(self) -> T:
```
turian, line 137: Do we want to call this from_0to1 and to_0to1? I just find it confusing each to time think through what "in range" means because it's kinda idiomatic and could mean either convention

src/ddspdrum/torchmodule.py
```diff
@@ -108,76 +85,63 @@ def npyforward(
...
-    def _update_modparameters(self) -> None:
-        for modparameter_id in self.modparameters:
-            self.modparameters[modparameter_id].set_value(
-                float(self.torchparameters[modparameter_id].numpy())
-            )
-
-    # In general, we should consider removing all the following
-    # since we don't want to cast to int and then back to Tensor
-    # Alternately, we should make a SynthParameter which is an nn.Parameter
-    # but decorated with ModParameter stuff.
```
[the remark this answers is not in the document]
turian: Yeah i like this

src/ddspdrum/torchmodule.py
```diff
@@ -108,76 +85,63 @@ def npyforward(
...
         Parameters
         ----------
         modparameter_id (str)  :   Id of the modparameter to return the value for
         """
-        self._update_modparameters()
-        return self.modparameters[modparameter_id].get_value_0to1()
+        return float(self.torchparameters[parameter_id].item())
 
-    def set_modparameter(self, modparameter_id: str, value: float):
+    def set_parameter(self, parameter_id: str, value: float):
```
turian, line 116: Maybe we should establish the convention that all these methods are either `_0to1` or `_human`/`_orig`??

turian: Don't have to tho

turian (approved):
This is good, we can merge it or tweak it a little, as you like

src/ddspdrum/torchmodule.py
jorshi: Going to leave this for now, I think it makes sense at this level that an end-user would be interacting with it that you would want to set a parameter in the human range by default, and then having the other setter explicitly marked set_parameter_0to1. Happy to change this in the future if you find it confusing.

jorshi:
Updated with some changes suggested by @turian and merged

PR #61, opened 2021-01-30
Fm vco

maxsolomonhenry:
Made `FmVCO` class, which basically has to process the modulation signal in a slightly different way. Typically the mod signal is applied in midi-space (log frequency), but FM operates on modulations in Hz space. Also changed the modulation depth to reflect the classic 'modulation index' of FM literature. It's a bit more intuitive this way.

Had to slightly refactor `VCO` to make this smooth.

Note, this is only work on the numpy modules. `torchmodule.py` would have to be updated accordingly.


src/ddspdrum/module.py
```diff
@@ -368,16 +368,21 @@ def _npyforward(self, mod_signal: np.array, phase: float = 0.0) -> np.ndarray:
...
 
+        # Store final phase.
         self.phase = cosine_argument[-1]
         return self.oscillator(cosine_argument)
 
+    def make_control_as_frequency(self, mod_signal: np.ndarray):
+        modulation = self.p("mod_depth") * mod_signal
+        control_as_midi = self.p("pitch") + modulation
+        return midi_to_hz(control_as_midi)
+
```
maxsolomonhenry, line 385: Factoring this section, as this is what `FmVCO` has to replace. Could be a better name for this method.

Effectively we're pivoting between various representations of instructions: (1) *midi* values, (2) *instantaneous frequency* (Hz), and (3) *phase*, which ultimately controls the oscillator.

jorshi: pitch_signal_to_frequency? I think what you have is also okay

src/ddspdrum/module.py
```diff
@@ -368,16 +368,21 @@ def _npyforward(self, mod_signal: np.array, phase: float = 0.0) -> np.ndarray:
```
turian, line 368: hmmmm so changing numpy functionality that has already been ported to torch (VCO), you should also port these changes to torch.

src/ddspdrum/module.py
```diff
@@ -368,16 +368,21 @@ def _npyforward(self, mod_signal: np.array, phase: float = 0.0) -> np.ndarray:
 
         """
 
-        assert (mod_signal >= 0).all() and (mod_signal <= 1).all()
+        assert (mod_signal >= -1).all() and (mod_signal <= 1).all()
```
turian, line 371: Can you explain why we need this change? Is it only FmVCO where mod_signal >= -1 and for all the others its >= 0?

src/ddspdrum/module.py
```diff
@@ -401,14 +406,51 @@ class SineVCO(VCO):
...
+        self,
+        midi_f0: float = 10.0,
+        mod_depth: float = 50.0,
+        phase: float = 0.0,
     ):
+
         super().__init__(midi_f0=midi_f0, mod_depth=mod_depth, phase=phase)
 
     def oscillator(self, argument):
         return np.cos(argument)
```
turian, line 420: Pretty cool.

src/ddspdrum/module.py
maxsolomonhenry: Frequency modulation adjusts the frequency both above and below the central frequency; that's how the theory works and how you get the nice central pitch percept. 

But I'm thinking that mod's should be `-1 < mod < 1` in general though. Makes for a more flexable synth.

src/ddspdrum/torchmodule.py
```diff
@@ -435,3 +439,60 @@ def __init__(
...
+        super().__init__(sample_rate=sample_rate, buffer_size=buffer_size)
+
+    def _forward(self, control_in: T, audio_in: T) -> T:
+        assert (control_in >= 0).all() and (control_in <= 1).all()
+
+        if (audio_in <= -1).any() or (audio_in >= 1).any():
+            normalize(audio_in)
+
+        audio_in = fix_length(audio_in, len(control_in))
+        return control_in * audio_in
```
maxsolomonhenry, line 498: This might be controversial, but needed to be addressed in both the npy and torch modules:

Before the VCA clipped both control and audio signals. Here, we assert the control is in the right range (unlike pitch mod, no reason for it to ever be negative; and no reason to be above 1). The audio is generally well-behaved, but sometimes can creep outside [-1, 1]. With most random synths this is not the case, but it's here as a fail safe. It shouldn't effect synth performance.

src/ddspdrum/module.py
turian: Okay cool

turian:
Cool great :) Do you want to add tests? If you're having trouble I can help

turian (approved):
This looks really good

Issue #66, opened 2021-02-06
ADSR one-shot should be a separate module


turian:
```

        # If sustain is "0" go to one-shot mode (moves through ADR sections).
        if note_on_duration == T(0):
            note_on_duration = self.p("attack") + self.p("decay")
```

I want to go back to my argument that having this special case makes it harder to generate random patches since this value 0 isn't really described correctly but the parameter range and curve

PR #67, opened 2021-02-08
ADSR remove one-shot

jorshi:
Removing the one-shot condition re issue #66 -- no separate module for now for one-shot mode since the envelope for that can be created by setting the parameters correctly in the ADSR as it is now. If we feel like we want to add a one-shot module in the future then we can do that.


maxsolomonhenry:
This looks great.

PR #68, opened 2021-02-10
Torch filters

jorshi:
Starting to make some differentiable filters. Moving average and FIR LPF are included in this PR. In the future I think we are going to want to compute convolutions in the frequency domain as opposed to doing it using torch conv1d as that will likely be more efficient, this is what Engel et al. did in the DDSP work. But I will leave that for a future change since this is working for now. Not including fuzzy tests since the numpy convolution operation and torch.nn.functional.conv1d operations aren't quite the same.

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -88,14 +88,7 @@ envelope = adsr.npyforward(note_on_duration)
 time_plot(envelope, adsr.sample_rate)
 # -
 
-# ### One-Shot Mode
-#
-# Alternately, you can specify a sustain time of "0" which will switch the
-# envelope to one-shot mode. In this case, the envelope moves through the entire
-# attack, decay, and release.
-
-envelope = adsr.npyforward(note_on_duration = 0)
-time_plot(envelope, adsr.sample_rate)
+# ## Oscillators
 
 # SineVCO test
 midi_f0 = 12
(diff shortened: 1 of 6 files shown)


src/ddspdrum/torchmodule.py
```diff
@@ -485,3 +485,146 @@ def _forward(self, control_in: T, audio_in: T) -> T:
...
+        audio (T)  :   audio samples to filter
+        """
+
+        impulse = self.windowed_sinc(self.p("cutoff"), self.p("length"))
+        impulse = torch.reshape(impulse, (1, 1, impulse.size()[0]))
+        audio_resized = torch.reshape(audio_in, (1, 1, audio_in.size()[0]))
+        y = nn.functional.conv1d(audio_resized, impulse, padding=int(self.p("length") / 2))
+        return y[0][0]
+
+    def windowed_sinc(self, cutoff: T, filter_length: T) -> T:
```
jorshi, line 545: I don't love the code in this method, but it was challenging to make this differentiable in a way that felt really clean. The big obstacle that I encountered in both of the filters was making the filter length differentiable.

tests/test_torchmodule.py
```diff
@@ -107,6 +107,7 @@ def test_TorchSineVCO(self):
         )
 
     def test_TorchFmVCO(self):
+        # Is this correct?
         numpymod = numpymodule.FmVCO()
         torchmod = torchmodule.TorchSineVCO()
```
jorshi, line 112: Not really part of this PR, but wanted to flag this. should this be comparing FmVCO and TorchFmVCO -- why is this test working?

tests/test_torchmodule.py
```diff
@@ -117,6 +118,18 @@ def test_TorchFmVCO(self):
             }
         )
 
+    # Fuzzy test doesn't work for filtering
```
jorshi, line 121: Going to pull this out.

src/ddspdrum/torchmodule.py
```diff
@@ -485,3 +485,146 @@ def _forward(self, control_in: T, audio_in: T) -> T:
 
         audio_in = fix_length(audio_in, len(control_in))
         return control_in * audio_in
+
+
+class TorchFIR(TorchSynthModule):
```
jorshi, line 490: Considering renaming this -- FIRLowPass or something similar that is more specific

tests/test_torchmodule.py
turian: hmmm @maxsolomonhenry ? thoughts?

src/ddspdrum/torchmodule.py
```diff
@@ -485,3 +485,154 @@ def _forward(self, control_in: T, audio_in: T) -> T:
...
+        TODO: Cutoff frequency modulation, if there is an efficient way to do it
+
+        Parameters
+        ----------
+
+        audio (T)  :   audio samples to filter
+        """
+
+        impulse = self.windowed_sinc(self.p("cutoff"), self.p("length"))
+        impulse = torch.reshape(impulse, (1, 1, impulse.size()[0]))
```
turian, line 540: reshape looses gradients I think? You probably want view?

Are you able to backprop through this?

src/ddspdrum/torchmodule.py
turian: Cool, might also be worth comparing to https://github.com/mravanelli/SincNet/blob/788fc3cd42e5c3172c5a8525b3a872eb7205c8f9/dnn_models.py and https://github.com/adefossez/julius

tests/test_torchmodule.py
maxsolomonhenry: Hah. Whoops. Would you mind tweaking it? Odd how it didn't seem to trigger any failure...

src/ddspdrum/torchmodule.py
jorshi: Nice, the julius code was helpful for cleaning up the windowed sinc implementation a bit. That FFT based convolution implementation is sweet, would be cool to explore that in the future.

tests/test_torchmodule.py
jorshi: Consider it tweaked. That is odd that it didn't trigger any failure?

src/ddspdrum/torchmodule.py
jorshi: Yes able to backprop through this -- is view better though?

jorshi: So I guess reshape will change the view if possible, otherwise it may copy the data -- so I think view is better here, made this change.

examples/examples.py
```diff
@@ -554,3 +554,112 @@ def stft_plot(signal, sample_rate=SAMPLE_RATE):
...
+# Create some noise to filter
+duration = 2
+noise = torch.tensor(np.random.rand(int(44100 * duration)) * 2 - 1).float()
+stft_plot(noise.detach().numpy())
+# -
+
+# **Moving Average Filter**
+#
+# A moving average filter is a simple finite impulse response (FIR) filter that calculates that value of a sample by taking the average of M input samples at a time. The filter_length defines how many samples M to include in the average.
+
```
maxsolomonhenry, line 573: As I mentioned in the fb chat, the output here looks like and sounds like comb filter and not an MA filter. Not sure about the guts of it; and maybe I'm off... but would expect an LP type behaviour.

I might be totally off base. Lmk.

PR #69, opened 2021-02-17
Running current examples on GPU

jorshi:
Tested on my local machine to see if I could get current modules in main to run on the GPU -- looks like it is working as far as I can tell

### DIFF
--- a/examples/examples.py
+++ b/examples/examples.py
@@ -411,8 +411,15 @@ ipd.Audio(kick, rate=sample_rate)
 # ## Torch examples
 
 
+# +
 import torch
 from ddspdrum.torchmodule import TorchADSR
+
+if torch.cuda.is_available():
+    device = "cuda:0"
+else:
+    device = "cpu"
+# -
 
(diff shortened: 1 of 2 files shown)


maxsolomonhenry:
I've got absolutely no issue with this PR in principle. Whenever you wanna pull it in, be my guest.

examples/examples.py
```diff
@@ -411,9 +411,16 @@ def stft_plot(signal, sample_rate=SAMPLE_RATE):
 # ## Torch examples
 
 
+# +
 import torch
 from ddspdrum.torchmodule import TorchADSR
 
+if torch.cuda.is_available():
+    device = "cuda:0"
```
turian, line 419: I'm not 100% sure about this, I think if you use "cuda" then it uses all your GPUs. We have an issue later to test it: https://github.com/turian/torchsynth/issues/71 so this is fine for now

turian (approved):
Looks fine to me so far. We'll improve this over time

PR #70, opened 2021-02-24
Torch SVF Filters

jorshi:
State variable filter implementation that seems to be working in Torch and computing gradients that can be used to optimize it.

### DIFF
--- a/src/ddspdrum/torchmodule.py
+++ b/src/ddspdrum/torchmodule.py
@@ -517,8 +517,12 @@ class TorchSVF(TorchSynthModule):
         **kwargs
     ):
         super().__init__(**kwargs)
+
+        # Set the filter type
+        self.mode = mode.lower()
+        assert mode in ['lpf', 'hpf', 'bpf', 'bsf']
+
         nyquist = self.sample_rate / 2.0
-        self.mode = mode
         self.add_parameters(
             [
                 TorchParameter(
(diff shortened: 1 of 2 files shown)


src/ddspdrum/torchmodule.py
```diff
@@ -485,3 +485,189 @@ def _forward(self, control_in: T, audio_in: T) -> T:
...
+            [
+                TorchParameter(
+                    value=cutoff,
+                    parameter_range=ParameterRange(5.0, nyquist, "log"),
+                    parameter_name="cutoff"
+                ),
+                TorchParameter(
+                    value=resonance,
+                    parameter_range=ParameterRange(0.01, 1000.0, "log"),
+                    parameter_name="resonance"
```
maxsolomonhenry, line 536: Should this lower limit be 0.5 ?

turian:
okay I just think we should be avoiding modules that are known to be slow.

src/ddspdrum/torchmodule.py
jorshi: So it really just needs to be greater than 0 -- I set it to 0.5 kind of arbitrarily in the numpy version -- 0.01 gives it quite a bit more room for less intense resonance and wider bandpasses / bandstops

jorshi:
@turian -- ya I understand that. This kind of filter is important to synthesis though so I think there is an argument to including it with the idea that maybe there is a way to make it faster at some point? Going to merge this and if we decide that we don't want / never use it we can pull it out.

PR #74, opened 2021-03-04
Torchport

jorshi:


src/ddspdrum/torchmodule.py
```diff
@@ -486,12 +529,164 @@ def _forward(self, control_in: T, audio_in: T) -> T:
...
+        ----------
+        modules (Dict): A dictionary of TorchSynthModule
+        """
+
+        for name in modules:
+            if not isinstance(modules[name], TorchSynthModule):
+                raise TypeError(f"{modules[name]} is not a TorchSynthModule")
+
+            if modules[name].sample_rate != self.sample_rate:
+                raise ValueError(f"{modules[name]} sample rate does not match")
```
jorshi, line 608: Should we be throwing an error here? Or perhaps just a warning and then set the sample rate of the child module being added to the sample rate of the Synth?

src/ddspdrum/torchmodule.py
```diff
@@ -486,12 +529,164 @@ def _forward(self, control_in: T, audio_in: T) -> T:
...
+            ratio: float = 0.25,
+            **kwargs
+    ):
+        super().__init__(**kwargs)
+        self.add_parameters(
+            [
+                TorchParameter(
+                    value=ratio,
+                    parameter_name="ratio",
+                    parameter_range=ParameterRange()
```
maxsolomonhenry, line 559: `(0, 1)` ?

src/ddspdrum/torchmodule.py
maxsolomonhenry: I think error, because if the sample rates aren't matching there's surely some bigger problem earlier in the chain. My $0.02

src/ddspdrum/torchmodule.py
```diff
@@ -486,12 +529,164 @@ def _forward(self, control_in: T, audio_in: T) -> T:
...
+        self.note_on_duration = T(note_on_duration)
+
+        # Setup a dummy global parameter module
+        global_params = TorchSynthModule(
+            self.sample_rate.item(),
+            self.buffer_size.item()
+        )
+        global_params.add_parameters([
+            TorchParameter(vco_ratio, "vco_ratio", ParameterRange())
+        ])
```
maxsolomonhenry, line 656: Cool. Can't help but wonder about other design options here. This works for sure. Just feels a little, I dunno, my spidey sense is tingling. If you feel likewise, great — if not, happy to leave it a move on.

src/ddspdrum/torchmodule.py
maxsolomonhenry: I think so, yeah (throw an error). Because if the rates aren't matching, then surely there's something off earlier in the code.

src/ddspdrum/torchmodule.py
```diff
@@ -486,12 +529,164 @@ def _forward(self, control_in: T, audio_in: T) -> T:
...
+            ]
+        )
+
+    def _forward(self, audio_in: T) -> T:
+        noise = self.noise_of_length(audio_in)
+        return util.crossfade(audio_in, noise, self.p("ratio"))
+
+    @staticmethod
+    def noise_of_length(audio_in: T) -> T:
+        return torch.rand_like(audio_in) * 2 - 1
```
turian, line 570: We'll fix this later: https://github.com/turian/torchsynth/issues/76

src/ddspdrum/torchmodule.py
turian: Yeah I prefer complain loudly and fail

turian: In general, I like assertions (or valueerror raises) and think we should use them liberally

src/ddspdrum/torchmodule.py
turian: @maxsolomonhenry can you explain the design question here?

src/ddspdrum/torchmodule.py
jorshi: These are default values for ParameterRange -- happy to explicitly add them though, might be helpful for readability

src/ddspdrum/torchmodule.py
jorshi: This creates the random values on the same device as audio_in, so if audio_in is on the GPU then this is equivalent to torch.rand_like(audio_in, device="cuda")

PR #79, opened 2021-03-05
Cleaner way of doing global synth params

jorshi:


src/ddspdrum/torchmodule.py
```diff
@@ -52,7 +52,7 @@ def _forward(self, *args: Any, **kwargs: Any) -> T:  # pragma: no cover
         """
         Each TorchSynthModule should override this.
         """
-        pass
+        raise NotImplementedError("Derived classes must override this method")
```
jorshi, line 55: Enforce implementation of _forward

src/ddspdrum/torchmodule.py
```diff
@@ -556,7 +556,7 @@ def __init__(
                 TorchParameter(
                     value=ratio,
                     parameter_name="ratio",
-                    parameter_range=ParameterRange()
+                    parameter_range=ParameterRange(0.0, 1.0)
```
jorshi, line 559: Being specific here -- this was from Max's comment in PR #74

src/ddspdrum/torchmodule.py
```diff
@@ -570,6 +570,22 @@ def noise_of_length(audio_in: T) -> T:
         return torch.rand_like(audio_in) * 2 - 1
 
 
+class TorchSynthParameters(TorchSynthModule):
```
jorshi, line 573: Open to alternative names for this

turian (approved):

PR #99, opened 2021-03-07
Add simple signal class

turian:
We now have a simple Signal class for 2D signals, with convenience methods batch_size and num_samples.

If we start typing everything with Signal as we port, and use the batch_size and num_samples convenience properties, we will avoid bugs like squeezing things to the wrong ndim


jorshi (approved):
Cool, I'm into it

PR #111, opened 2021-03-08
Mechanical rename of everything

turian:


torchsynth/module.py
```diff
@@ -150,7 +62,7 @@ def seconds_to_samples(self, seconds: T) -> T:
 
     def _forward(self, *args: Any, **kwargs: Any) -> Signal:  # pragma: no cover
         """
-        Each TorchSynthModule should override this.
+        Each TorchSynthModule0Ddeprecated should override this.
```
jorshi, line 65: why is this TorchSynthModule0Ddeprecated?

torchsynth/module.py
```diff
@@ -600,14 +512,14 @@ def _forward(self, control_in: Signal, audio_in: Signal) -> Signal:
...
-class TorchNoise(TorchSynthModule1D):
+class TorchNoise(TorchSynthModule):
     """
     Adds noise to a signal
 
     Parameters
     ----------
     ratio (float): mix ratio between the incoming signal and the produced noise
-    **kwargs: see TorchSynthModule
+    **kwargs: see TorchSynthModule0Ddeprecated
```
jorshi, line 522: Same as above

turian: mechanical error. I fixed in later PR but push here anyway.

torchsynth/module.py
turian: mechanical error. I fixed in later PR but push here anyway.

PR #112, opened 2021-03-08
Globals

turian:
We not just pass around SynthGlobals everywhere. This makes our lives easier, as you will see in the next PR. Merge #111 first

PR #113, opened 2021-03-08
Parameter refactor

turian:
I want to be able to construct modules without necessarily providing any values to them. This cleans up the code *a lot*. Review and merge #112 first

### DIFF
--- a/torchsynth/module.py
+++ b/torchsynth/module.py
@@ -241,9 +241,6 @@ class TorchADSR(TorchSynthModule):
         ),
     ]
 
-    def __init__(self, synthglobals: TorchSynthGlobals, **kwargs: Dict[str, T]):
-        super().__init__(synthglobals, **kwargs)
-
     def _forward(self, note_on_duration: T) -> Signal:
         """Generate an ADSR envelope.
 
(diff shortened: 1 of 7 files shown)


torchsynth/module.py
```diff
@@ -377,36 +344,38 @@ class TorchVCO(TorchSynthModule1D):
...
+        ModuleParameterRange(
+            0.0, 127.0, name="midi_f0", description="pitch value in 'midi' (69 = 440Hz)"
+        ),
+        ModuleParameterRange(
+            0.0,
+            127.0,
+            name="mod_depth",
+            description="depth of the pitch modulation in semitones",
+        ),
+        # TODO: I think phase should go in here??
```
jorshi, line 361: I don't think we need phase as a ranged parameter. It's sort of a weird one -- it's important to keep track of to avoid clicks on consecutive calls to forward, but varying it like a regular synth parameter isn't going to do that much perceptually. I think pretty much always we are just going to want it to be initialized to zero on construction. Interested in @maxsolomonhenry thoughts.

jorshi (commented):
sweet - really into this change as well. clean!

torchsynth/module.py
turian: oki then we leave it as is

torchsynth/module.py
```diff
@@ -133,13 +78,44 @@ def __init__(
...
+                        parameter_name=parameter_range.name,
+                        data=torch.rand((self.synthglobals.batch_size,)),
+                        parameter_range=parameter_range,
+                    )
+                    for parameter_range in self.parameter_ranges
+                ]
+            )
+            if kwargs:
+                for name, data in kwargs.items():
+                    self.set_parameter(name, data)
```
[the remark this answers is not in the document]
turian: p cool

torchsynth/parameter.py
```diff
@@ -108,10 +119,11 @@ class ModuleParameter(nn.Parameter):
 
     def __new__(
         cls,
-        value: T = None,
+        # TODO: REMOVEME
+        value: Optional[T] = None,
```
jorshi, line 123: Why remove value?

torchsynth/parameter.py
```diff
@@ -148,6 +160,7 @@ def __repr__(self):
     #        self.data == other.data
     #    )
 
+    # TODO: Move to ModuleRange
```
jorshi, line 163: What's the idea here?

torchsynth/parameter.py
turian: If we want defaults that aren't 0.5, they should go into moduleparameterrange

PR #116, opened 2021-03-09
TorchDrum module to 2D

jorshi:


torchsynth/module.py
```diff
@@ -651,6 +648,31 @@ def _forward(self, *args: Any, **kwargs: Any) -> T:
         raise RuntimeError("TorchSynthParameters cannot be called")
 
 
+class TorchIdentity(TorchSynthModule):
```
jorshi, line 651: What is this going to be used for?

turian: For when you want to ignore a particular oscillator

turian: I was trying to replicate the old examples.py

PR #120, opened 2021-03-09
Facilitate various note_on_dur

maxsolomonhenry:
It's a little hairy but the adsr's will now work with various note_on_dur sizes. Needs to be refactored once note_on_dur is a parameter.

Also, "attack" behaviour isn't perfect, as it always goes up to 1 rather than some fractional value if `hold_on_dur` is less than the attack amount. That's ok though, this is a kind of self-normalization.

### DIFF
--- a/torchsynth/module.py
+++ b/torchsynth/module.py
@@ -10,7 +10,6 @@ import torch.tensor as T
 
 import torchsynth.util as util
 from torchsynth.defaults import DEFAULT_BUFFER_SIZE, DEFAULT_SAMPLE_RATE, EPSILON
-
 from torchsynth.parameter import ModuleParameter, ModuleParameterRange
 from torchsynth.signal import Signal
 
--- a/torchsynth/module.py
+++ b/torchsynth/module.py
@@ -306,6 +306,13 @@ class TorchADSR(TorchSynthModule):
         ramp = (ramp + EPSILON) / (duration_[:, None] + EPSILON)
         ramp = torch.minimum(ramp, T(1.0))
 
+        """
+        The following is a workaround. In inverse mode, a ramp with 0 duration 
+        (that is all 1's) becomes all 0's, which is a problem for the 
+        ultimate calculation of the ADSR signal (a * d * r => 0's). So this 
+        replaces only rows who sum to 0 (i.e., all components are zero).
+        """
+
         if inverse:
             ramp = 1.0 - ramp
             ramp[torch.sum(ramp, axis=1) == 0] = 1.0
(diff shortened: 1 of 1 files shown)


torchsynth/module.py
```diff
@@ -303,24 +303,29 @@ def _ramp(self, start, duration: T, inverse: bool = False) -> Signal:
...
         ramp = ramp - start_[:, None]
         ramp = torch.maximum(ramp, T(0.0))
-        ramp = ramp / duration_[:, None]
+        ramp = (ramp + EPSILON) / (duration_[:, None] + EPSILON)
         ramp = torch.minimum(ramp, T(1.0))
 
         if inverse:
-            ramp = 1 - ramp
+            ramp = 1.0 - ramp
+            ramp[torch.sum(ramp, axis=1) == 0] = 1.0
```
maxsolomonhenry, line 311: This is a funny workaround. In inverse mode, a ramp with 0 duration (that is all 1's) becomes all 0's, which is a problem for the ultimate calculation of the adsr signal (a * d * r => 0's). So this replaces only rows who sum to 0 (i.e., all components are zero).

torchsynth/module.py
```diff
@@ -303,24 +303,29 @@ def _ramp(self, start, duration: T, inverse: bool = False) -> Signal:
...
-        attack = self.p("attack")
+    def make_attack(self, note_on_duration) -> Signal:
+        attack = torch.minimum(self.p("attack"), note_on_duration)
         return self._ramp(torch.zeros(self.batch_size, device=attack.device), attack)
 
-    def make_decay(self) -> Signal:
+    def make_decay(self, note_on_duration) -> Signal:
+        attack = torch.minimum(self.p("attack"), note_on_duration)
+        decay = torch.maximum(note_on_duration - self.p("attack"), T([0.0]))
+        decay = torch.minimum(decay, self.p("decay"))
```
maxsolomonhenry, line 325: Calculations to accommodate a note_hold_on value that might be cutting into the attack/decay phase of the envelope. Not sure how this will differentiate.

torchsynth/module.py
```diff
@@ -303,24 +303,38 @@ def _ramp(self, start, duration: T, inverse: bool = False) -> Signal:
...
 
-    def make_attack(self) -> Signal:
-        attack = self.p("attack")
+    def make_attack(self, note_on_duration) -> Signal:
+        attack = torch.minimum(self.p("attack"), note_on_duration)
         return self._ramp(torch.zeros(self.batch_size, device=attack.device), attack)
 
-    def make_decay(self) -> Signal:
+    def make_decay(self, note_on_duration) -> Signal:
+        attack = torch.minimum(self.p("attack"), note_on_duration)
```
jorshi, line 330: This is being calculated twice -- wondering if we can calculate the attack duration and decay duration outside these functions and pass the attack time and decay time into these functions instead?

maxsolomonhenry: Fixed, thanks.

jorshi (approved):

PR #121, opened 2021-03-09
Drum GPU

jorshi:
Builds on #116 and makes changes for TorchDrum to run on GPU

### DIFF
--- a/torchsynth/module.py
+++ b/torchsynth/module.py
@@ -377,7 +377,9 @@ class TorchVCO(TorchSynthModule):
         **kwargs: Dict[str, T],
     ):
         super().__init__(synthglobals, **kwargs)
-        self.phase = self.p("initial_phase").unsqueeze(1)
+        self.phase = nn.Parameter(
+            data=self.get_parameter("initial_phase").unsqueeze(1), requires_grad=False
+        )
 
     def _forward(self, mod_signal: Signal) -> Signal:
         """
@@ -406,7 +408,7 @@ class TorchVCO(TorchSynthModule):
         assert (mod_signal >= -1).all() and (mod_signal <= 1).all()
         control_as_frequency = self.make_control_as_frequency(mod_signal)
         cosine_argument = self.make_argument(control_as_frequency) + self.phase
-        self.phase = cosine_argument[:, -1]
+        self.phase.data = cosine_argument[:, -1]
         output = self.oscillator(cosine_argument)
         return output.as_subclass(Signal)
 
--- a/torchsynth/module.py
+++ b/torchsynth/module.py
@@ -393,6 +393,10 @@ class TorchVCO(TorchSynthModule):
         **kwargs: Dict[str, T],
     ):
         super().__init__(synthglobals, **kwargs)
+
+        # TODO: Currently making phase a parameter with no grad
+        # Is there a way to do this without making it a param?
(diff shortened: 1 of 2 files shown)


torchsynth/module.py
```diff
@@ -376,27 +379,23 @@ class TorchVCO(TorchSynthModule):
...
-        else:
-            # Create initial phase of zeros like the parameters
-            self.phase = nn.Parameter(
-                data=torch.zeros((self.batch_size,)).unsqueeze(1), requires_grad=False
-            )
-
-        assert self.phase.shape[0] == self.batch_size
+        self.phase = nn.Parameter(
+            data=self.get_parameter("initial_phase"), requires_grad=False
+        )
```
jorshi, line 398: Not sure if there is a better way to do this -- making it a parameter makes sure that it gets transferred to the GPU when the module is transferred

turian: hmm no don't think so

we would be tuning initial phase but not the phase if we rerun this several times is my understanding.

so this should just be a like:
self.phase = T(self.get_parameter("initial_phase"), requires_grad, device=device)

See https://towardsdatascience.com/7-tips-for-squeezing-maximum-performance-from-pytorch-ca4a40951259 Construct tensors directly on GPUs

torchsynth/module.py
```diff
@@ -405,8 +423,9 @@ def _forward(self, mod_signal: Signal) -> Signal:
 
         assert (mod_signal >= -1).all() and (mod_signal <= 1).all()
         control_as_frequency = self.make_control_as_frequency(mod_signal)
-        cosine_argument = self.make_argument(control_as_frequency) + self.phase
-        self.phase = cosine_argument[:, -1]
+        cosine_argument = self.make_argument(control_as_frequency)
+        cosine_argument += self.phase.unsqueeze(1)
```
turian, line 427: Later but not now we should figure out if unsqueeze is a view (fast) or clone (slow)

torchsynth/module.py
```diff
@@ -757,8 +776,10 @@ def __init__(
         # with the mindset that if you are trying to learn to
         # synthesize a sound, you won't be adjusting the note_on_duration.
         # However, this is easily changed if desired.
-        # TODO: Fix this later
-        self.note_on_duration = T([note_on_duration] * synthglobals.batch_size)
+        # TODO: Fix this later -- should it actually be a parameter?
```
turian, line 779: Yes link to this issue in code https://github.com/turian/torchsynth/issues/117

turian: Not only should it be a parameter, it should be a TorchKnob

torchsynth/module.py
jorshi: the problem is that the parameters haven't been transferred to the GPU by this point. if you call:
```
vco = TorchSineVCO().to(device)
```
the constructor is called and self.phase gets created on the CPU. Then all the parameters get moved onto the GPU on the `to(device)` call, but self.phase stays on the CPU, which is a problem. By making self.phase a nn.Parameter, then it also gets moved to the GPU on `to(device)`

torchsynth/module.py
jorshi: it's a view: "The returned tensor shares the same underlying data with this tensor."

PR #129, opened 2021-03-10
Manual setting of parameters

turian:
If Synth.forward() is given an int, that int is used to generate the synth parameters as the random seed.


torchsynth/module.py
```diff
@@ -354,7 +354,12 @@ def __init__(
         # TODO: Make sure this is on GPU
         # https://pytorch-lightning.readthedocs.io/en/latest/advanced/multi_gpu.html#init-tensors-using-type-as-and-register-buffer
         # Do we want to detach clone?
-        self.phase = self.get_parameter("initial_phase").detach().clone()
+        # Do we want this persistent?
```
jorshi, line 357: So persistence would store this in the state_dict -- what would be the benefit of that? Saving and loading a synth? I don't think that keeping these values is important as long as it is always initialized with the initialize phase parameter values.

PR #135, opened 2021-03-10
Actually lightning

turian:
Timeline on 2 GPUs for 1B

image


jorshi (approved):

examples/lightningsynth.py
```diff
@@ -1,29 +1,96 @@
...
+        self.num_batches = num_batches
+
+    def __getitem__(self, idx):
+        return idx
+
+    def __len__(self):
+        return self.num_batches
+
+
+synth1B = batch_idx_dataset(1024 * 1024 * 1024 // BATCH_SIZE)
```
jorshi, line 66: Do we want this to be something a bit closer to a billion? I guess it's not a big issue but this is 73M larger than 1B.

jorshi:
also maybe add lightning_logs/ to gitignore

PR #137, opened 2021-03-10
Cryptographically secure pseudo-random numbers!

turian:
With unit tests, so the results are replicable.

We should profile this but I still think pow and sin will be the biggest culprits.

### DIFF
--- a/examples/lightningsynth.py
+++ b/examples/lightningsynth.py
@@ -1,3 +1,4 @@
+#!/usr/bin/env python3
 # -*- coding: utf-8 -*-
 # # lightningsynth
 #
@@ -9,18 +10,20 @@
 # # !pip install git+https://github.com/turian/torchsynth.git@lightning-synth
 # ```
 
-from torchsynth.synth import Voice
-from torchsynth.globals import SynthGlobals
-
+import torch
+import torch.tensor as T
 from tqdm.auto import tqdm
 
-import torch.tensor as T
-import torch
+from torchsynth.globals import SynthGlobals
+from torchsynth.synth import Voice
 
 synthglobals = SynthGlobals(batch_size=T(256))
-voice = Voice(synthglobals).cuda()
+voice = Voice(synthglobals)
+
+if torch.cuda.is_available():
+    voice.cuda()
 
 voice.eval()
 with torch.no_grad():
     for i in tqdm(range(1000)):
-        voice()
+        voice(i)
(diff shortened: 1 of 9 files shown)


torchsynth/synth.py
```diff
@@ -51,68 +54,90 @@ def buffer_size(self) -> T:
...
-                "note_on": NoteOnButton(synthglobals),
-                "pitch_adsr": ADSR(synthglobals),
-                "amp_adsr": ADSR(synthglobals),
-                "vco_1": SineVCO(synthglobals),
-                "vco_2": SquareSawVCO(synthglobals),
-                "noise": Noise(synthglobals),
-                "vca": VCA(synthglobals),
-                "vca_ratio": CrossfadeKnob(synthglobals),
-            }
+            [
```
turian, line 128: This is a list now to ensure all the modules are in deterministic order

torchsynth/module.py
```diff
@@ -354,7 +354,12 @@ def __init__(
         # TODO: Make sure this is on GPU
         # https://pytorch-lightning.readthedocs.io/en/latest/advanced/multi_gpu.html#init-tensors-using-type-as-and-register-buffer
         # Do we want to detach clone?
-        self.phase = self.get_parameter("initial_phase").detach().clone()
+        # Do we want this persistent?
+        # https://pytorch.org/docs/stable/generated/torch.nn.Module.html#torch.nn.Module.register_buffer
```
jorshi, line 358: Commented on this in other PR, but I don't think we need it persistent. If it is then it basically allows us to save and load a synth a keep these phase values right?

torchsynth/synth.py
```diff
@@ -51,68 +54,91 @@ def buffer_size(self) -> T:
...
+            self.randomize(seed=batch_idx)
+        return self._forward(*args, **kwargs)
+
+    def randomize(self, seed: Optional[int]):
+        """
+        Randomize all parameters
+        """
+        if seed:
+            # Profile to make sure this isn't too slow?
+            mt19937_gen = csprng.create_mt19937_generator(seed)
```
jorshi, line 111: cool -- I like that for extracting specific batches in a deterministic way

jorshi (approved):

examples/lightningsynth.py
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env python3
```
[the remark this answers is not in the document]
turian: maybe not we're still getting lightning working

torchsynth/module.py
turian: Okay make a PR to fix?

jorshi: Will do

PR #142, opened 2021-03-10
Lightning old

turian:
Stupid version of lightning that involves dummy fit


torchsynth/synth.py
```diff
@@ -102,6 +102,25 @@ def forward(
             self.randomize(seed=batch_idx)
         return self._forward(*args, **kwargs)
 
+    # For lightning
+    def training_step(self, batch, batch_idx):
+        print("batch_idx", batch_idx)
+        import sys
```
jorshi, line 108: does import sys have to be here?

turian: will clean

PR #146, opened 2021-03-11
Param range

jorshi:
Parameter ranging a bit more versatile with defined values and symmetric ranges.


maxsolomonhenry:
Looks great. High level comment. I like the use of a `nyquist` variable,for example,  for parameter scaling. It would be nice to use this consistently. Sometimes I see Sr // 2, sometimes Nyquist. Just a style thing.

PR #148, opened 2021-03-11
Deterministic noise

turian:
Each module now has a seed, that it can use for *cryptographically secure PRNG*

### DIFF
--- a/torchsynth/synth.py
+++ b/torchsynth/synth.py
@@ -116,7 +116,7 @@ class AbstractSynth(LightningModule):
         """
         Randomize all parameters
         """
-        if seed:
+        if seed is not None:
             # Profile to make sure this isn't too slow?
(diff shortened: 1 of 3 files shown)


torchsynth/synth.py
```diff
@@ -112,7 +112,7 @@ def test_step(self, batch, batch_idx):
         # We just return 0, which lightning accumulates as the test error
         return T(0.0, device=self.device)
 
-    def randomize(self, seed: Optional[int]):
+    def randomize(self, seed: Optional[int] = None):
```
jorshi, line 115: Noticing the `if seed` statement below -- it should be updated to `if seed is not None:`

turian: oops will fix

turian: fixed

jorshi (approved):

PR #153, opened 2021-03-12
Lightning callback

jorshi:
Running synth in torch lightning callback now -- works on multi-gpu setups


torchsynth/synth.py
```diff
@@ -105,11 +105,6 @@ def forward(
 
     # For lightning
     def test_step(self, batch, batch_idx):
```
turian, line 107: Make a note here that this is boilerplate for lightning

examples/lightningsynth.py
```diff
@@ -63,6 +65,20 @@ def __len__(self):
         return self.num_batches
 
 
+class TorchSynthCallback(pl.Callback):
```
turian, line 68: Yeah this is cool, so our API could be that you pass in an optional function, and that function does whatever it wants with this batch. That function also gets the batch index as a param

turian: We can do that later

turian (approved):

PR #155, opened 2021-03-13
Keyboard Module

jorshi:
- Updated the note on button to a Keyboard module that contains parameters for midi f0 and the note duration. This can be 'triggered' and it returns the midi_f0 and duration like triggering a keyboard. 
- VCOs now accept midi_f0 in the forward call and have a tuning parameter instead of having the midi_f0 as a parameter. 
- Updated Synth with a set_parameters method that can be used to set as many or little parameters at once
- Removed Identity -- more on that in comment in code


torchsynth/module.py
```diff
@@ -544,15 +548,6 @@ def noise_of_length(self, audio_in: Signal) -> Signal:
             return torch.rand_like(audio_in) * 2 - 1
 
 
-class Identity(SynthModule):
```
jorshi, line 547: So I don't think this was working the way we intended it to. In the case of a VCO in Voice, we would be passing through the pitch modulation envelope which would then be summed with the first VCO. This would offset the signal in the shape of the ADSR. I thought about adding a module that just returns zeros. But there was some trouble with that getting created on the right device. So, I vote we remove this and don't disable modules -- or if we want to then we can come up with another way to have variable architectures?

turian: okay to remove

torchsynth/module.py
```diff
@@ -568,14 +563,20 @@ class CrossfadeKnob(SynthModule):
     ]
 
 
-class NoteOnButton(SynthModule):
+class Keyboard(SynthModule):
```
turian, line 566: Call it MonophonicKeyboard?

turian: Close this issue https://github.com/turian/torchsynth/issues/117 when you merge

jorshi: Sounds good

PR #156, opened 2021-03-13
Four Operator FM Synth

jorshi:


torchsynth/module.py
```diff
@@ -455,12 +455,22 @@ class TorchFmVCO(VCO):
...
+        description="depth of the fm modulation",
+    )
+
     def make_control_as_frequency(self, midi_f0: T, mod_signal: Signal) -> Signal:
         # Compute modulation in Hz space (rather than midi-space).
         f0_hz = util.midi_to_hz(midi_f0 + self.p("tuning")).unsqueeze(1)
         fm_depth = self.p("mod_depth").unsqueeze(1) * f0_hz
         modulation_hz = fm_depth * mod_signal
-        return torch.clamp(f0_hz + modulation_hz, 0.0, self.nyquist)
+        return f0_hz + modulation_hz
```
jorshi, line 472: Turns out aliasing is feature of FM

examples/examples.py
```diff
@@ -226,7 +226,7 @@ def stft_plot(signal, sample_rate=DEFAULT_SAMPLE_RATE):
 
 # ## Oscillators
 #
-# There are several types of oscillators and sound generators available. Oscillators that can be controlled by an external signal are called voltage-coltrolled oscillators (VCOs) in the analog world and we adpot a similar approach here; oscillators accept an input control signal and produce audio output. We have a simple sine oscilator:`SineVCO`, a square/saw oscillator: `SquareSawVCO`, and an FM oscillator: `TorchFmVCO`. There is also a white noise generator: `Noise`.
+# There are several types of oscillators and sound generators available. Oscillators that can be controlled by an external signal are called voltage-coltrolled oscillators (VCOs) in the analog world and we adpot a similar approach here; oscillators accept an input control signal and produce audio output. We have a simple sine oscilator:`SineVCO`, a square/saw oscillator: `SquareSawVCO`, and an FM oscillator: `FmVCO`. There is also a white noise generator: `Noise`.
```
turian, line 229: Line too long, fix manually or use command
```
!}fmt
```
from within vi

torchsynth/module.py
turian: Make that a comment, maybe with some explanation of where you've seen this

torchsynth/synth.py
```diff
@@ -163,9 +165,178 @@ def _forward(self) -> T:
...
+            ]
+        )
+
+    def _forward(self, midi_f0: T, duration: T, modulation: Signal) -> Signal:
+        output = self.osc(midi_f0, modulation)
+        env = self.env(duration)
+        return self.amp(env, output)
+
+
+class FmAlgorithmKnob(SynthModule):
```
turian, line 200: Should be in module.py?

torchsynth/synth.py
```diff
@@ -163,9 +165,178 @@ def _forward(self) -> T:
...
+        )
+
+        # Add the operators
+        self.op1 = FmOperator(synthglobals)
+        self.op2 = FmOperator(synthglobals)
+        self.op3 = FmOperator(synthglobals)
+        self.op4 = FmOperator(synthglobals)
+
+        # Algorithm layouts - 11 different layouts from Ableton's Operator
+
```
turian, line 242: This is super confusing all these floats, can you give a little more detail here

torchsynth/synth.py
```diff
@@ -163,9 +165,178 @@ def _forward(self) -> T:
...
+    def _forward(self, midi_f0: T, duration: T, modulation: Signal) -> Signal:
+        output = self.osc(midi_f0, modulation)
+        env = self.env(duration)
+        return self.amp(env, output)
+
+
+class FmAlgorithmKnob(SynthModule):
+    """
+    A knob for transitioning between FM algorithms
```
turian, line 202: What are FM algorithms?

turian: (As a code comment)

torchsynth/synth.py
```diff
@@ -163,9 +165,178 @@ def _forward(self) -> T:
...
+
+    def _forward(self) -> T:
+
+        # Trigger keyboard
+        midi_f0, note_on_duration = self.keyboard()
+
+        # Determine algorithm mix
+        algorithm = self.algorithm.p("algorithm")
+        self.lower = torch.floor(algorithm).long()
+        self.upper = torch.ceil(algorithm).long()
```
turian, line 293: okay lol this totally isn't differentiable. please make an issue for it

torchsynth/module.py
```diff
@@ -632,4 +646,62 @@ def forward(self) -> Tuple[T, T]:
...
+
+    def forward(self) -> T:
+        """
+        Return all the fader values in a tensor with shape (batch_size, n_faders)
+        """
+        params = torch.stack([self.p(p) for p in self.torchparameters])
+
+        # If the sum of faders is greater then 1.0 then normalize
+        summed = torch.sum(params, dim=0)
+        return torch.where(summed > 1.0, params / summed, params).T
```
jorshi, line 690: Is this the right way to think about this? I want to allow for the sum of the parameters to be between 0 and 1, but never greater than 1 while also preserving the relationship between the parameters. Not sure if this is going to lead to weird behavior though. Is this confusing?

torchsynth/parameter.py
```diff
@@ -63,8 +63,8 @@ def from_0to1(self, normalized: T) -> T:
         normalized (T)     : value within [0,1] range to convert to range defined by
             minimum and maximum
         """
-        assert torch.all(0.0 <= normalized)
-        assert torch.all(normalized <= 1.0)
+        # assert torch.all(0.0 <= normalized)
+        # assert torch.all(normalized <= 1.0)
```
jorshi, line 67: Uncomment

torchsynth/module.py
```diff
@@ -455,12 +451,30 @@ class TorchFmVCO(VCO):
         modulation_index = frequency_deviation / modulation_frequency
     """
 
+    parameter_ranges: List[ModuleParameterRange] = VCO.parameter_ranges.copy()
+    parameter_ranges[0] = ModuleParameterRange(
```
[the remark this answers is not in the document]
jorshi: Add as comment in code

torchsynth/module.py
```diff
@@ -455,12 +451,36 @@ class TorchFmVCO(VCO):
         modulation_index = frequency_deviation / modulation_frequency
     """
 
+    parameter_ranges: List[ModuleParameterRange] = VCO.parameter_ranges.copy()
```
turian, line 454: Do we need this? If we inherit from VCO?

torchsynth/module.py
turian: idk? why not normalize all of them?

maxsolomonhenry (commented):
Ok, it's taken me a while to get my thoughts together on this. I mean, it's totally rad. There's absolutely no denying that. 

Ultimately it comes down to it being kindof half-way between generalizable/learnable (i.e., being able to make claims about exhaustively searching the FM-space), and also being necessarily arbitrary and limited in its approach. In the process, we're packing each voice with a lot of complex computation, and that will certainly eat into the goal of the synth being fast.

I'm reminded of the idea that FM calculations are fairly cheap, which is why the DX7 took off the way it did in the early days of digital synthesis. So I think the real issue is not the FM per se but the modulation routing.

Here, our hands are tied a bit: we have to make an arbitrary choice about number of operators, and even within that, we have to make an arbitrary choice about the kinds of inter-modulation algorithms we can accommodate in a learnable way. @jorshi I believe you described a factorial type approach.

It certainly makes sense to go with a four operator FM voice, which is popular enough (though the DX7 has 6 operators). The question, I guess, is this: can this modulation routing account for *any routing algorithm* between the four oscillators? Or, at least, any of the popular ones being used?

If we could say that (and, maybe we can right now), then I think we'd be in great shape. Because we are already using the arbitrary architecture of the 2-oscillator subtractive synth. The fact that the architecture is a *prior* is its strength, I think. So likewise here, we should be using algorithms that are common, that tend to get used.

But yeah, like, if we're going to have this voice in there, let's make sure it can do everything we need it to do so it's worth the computation. My $0.02.

torchsynth/module.py
```diff
@@ -400,10 +400,6 @@ def _forward(self, midi_f0: T, mod_signal: Signal) -> Signal:
         assert (mod_signal >= -1).all() and (mod_signal <= 1).all()
 
         control_as_frequency = self.make_control_as_frequency(midi_f0, mod_signal)
-        assert (control_as_frequency >= 0).all() and (
-            control_as_frequency <= self.nyquist
-        ).all()
```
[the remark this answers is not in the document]
jorshi: Sure, that's a good idea

torchsynth/module.py
jorshi: So we want to be able to have the sum of the parameters be any where between 0 and 1. If we normalize them all then the sum will always be 1. It's like having a mixing board that allows you to have a mixture level up to 1, and then when you exceed that it automatically balances the mixture to preserve the ratio between the signals without going above 1. @maxsolomonhenry any thoughts about this?

maxsolomonhenry: It would be consistent with that we're doing in the noise module, for example. Maybe not the most intuitive for usability, but lets us make the most of the [-1, 1] sample range. 

The rub with either is that we lose any meaningful variability in volume. But since we're batch normalizing in a lot of the eventual learning, I dunno it's probably fine.

torchsynth/synth.py
```diff
@@ -163,9 +173,111 @@ def _forward(self) -> T:
...
+            [
+                ("keyboard", MonophonicKeyboard(synthglobals)),
+                ("modulation", FmControl(synthglobals)),
+                ("op2_inputs", NormalizedFaderBank(synthglobals, [0.5])),
+                ("op3_inputs", NormalizedFaderBank(synthglobals, [0.25, 0.5])),
+                ("op4_inputs", NormalizedFaderBank(synthglobals, [0.25, 0.33, 0.5])),
+                ("mixer", SoftModeSelector(synthglobals, n_modes=4)),
+            ]
+        )
+
```
turian, line 229: So we really haven't thought this thru much. i.e. nesting Synths.

I'm concerned here that:
a) randomize won't hit all the sub-synth parameters
b) listing out all the parameter ranges will be broken

jorshi: We're definitely going to want to have stuff nest IMO. PyTorch is quite smart about nesting modules and making sure that parameters are organized nicely within modules -- I tested pretty thoroughly, but would be happy to do some automated stuff on this as well.

examples/examples.py
```diff
@@ -588,6 +607,59 @@ def stft_plot(signal, sample_rate=DEFAULT_SAMPLE_RATE):
...
+# +
+from torchsynth.synth import FmVoice
+
+fm = FmVoice(synthglobals1).to(device)
+fm.set_parameters(
+    {
+        ("keyboard", "midi_f0"): T([60.0]),
+        ("op1", "osc", "ratio"): T([1.0]),
+        ("op2", "osc", "ratio"): T([1.0]),
+        ("op2", "osc", "mod_depth"): T([1.0]),
```
jorshi, line 653: @turian here is an example of how parameters work with Synth nesting. The nested module is called op1, so you can refer to the specific parameter as op1 -> osc -> ratio or op2 -> osc -> ratio

torchsynth/synth.py
```diff
@@ -163,9 +173,111 @@ def _forward(self) -> T:
...
+        # Register children SynthModules
+        self.add_synth_modules(
+            [
+                ("keyboard", MonophonicKeyboard(synthglobals)),
+                ("modulation", FmControl(synthglobals)),
+                ("op2_inputs", NormalizedFaderBank(synthglobals, [0.5])),
+                ("op3_inputs", NormalizedFaderBank(synthglobals, [0.25, 0.5])),
+                ("op4_inputs", NormalizedFaderBank(synthglobals, [0.25, 0.33, 0.5])),
+                ("mixer", SoftModeSelector(synthglobals, n_modes=4)),
```
jorshi, line 226: This should be replaced with the new AudioMixer

torchsynth/synth.py
```diff
@@ -163,9 +173,111 @@ def _forward(self) -> T:
...
+
+    def __init__(self, synthglobals: SynthGlobals, *args, **kwargs):
+        AbstractSynth.__init__(self, synthglobals=synthglobals, *args, **kwargs)
+
+        # Register children SynthModules
+        self.add_synth_modules(
+            [
+                ("keyboard", MonophonicKeyboard(synthglobals)),
+                ("modulation", FmControl(synthglobals)),
+                ("op2_inputs", NormalizedFaderBank(synthglobals, [0.5])),
```
jorshi, line 223: NormalizedFaderBank can be replaced with ModulationMixer

jorshi:
Closing this and lets revisit this one as things have changed in main quite a bit

Issue #165, opened 2021-03-14
KL weighting for SoftModeSelector


turian:
We want to sample SoftModeSelector params in a way that maximize KL divergence from uniform.

PR #166, opened 2021-03-14
SoftModeSelector

turian:


torchsynth/module.py
```diff
@@ -588,3 +588,41 @@ class MonophonicKeyboard(SynthModule):
...
+            ModuleParameterRange(
+                0.0,
+                1.0,
+                name=f"mode{i}weight",
+                description=f"mode{i} weight, before normalization",
+            )
+            for i in range(n_modes)
+        ]
+        super().__init__(synthglobals, **kwargs)
+        self.softmax = torch.nn.Softmax(dim=0)
```
jorshi, line 619: Don't think we need softmax any more?

torchsynth/module.py
```diff
@@ -588,3 +588,41 @@ class MonophonicKeyboard(SynthModule):
...
+
+    def forward(self) -> Tuple[T, T]:
+        """
+        Normalize all mode weights so they sum to 1.0
+        """
+        # Is this tensor creation slow?
+        # But usually parameter stuff is not the bottleneck
+        params = torch.stack([p.data for p in self.torchparameters.values()])
+        return params / torch.sum(params, dim=0)
```
jorshi, line 628: Very cool, so idea would be that the mode with the largest value gets selected?

torchsynth/module.py
turian: REmoving

torchsynth/module.py
turian: Actually no, right now https://github.com/turian/torchsynth/issues/165

jorshi (approved):

PR #194, opened 2021-03-20
SineLFO

jorshi:


torchsynth/module.py
```diff
@@ -544,6 +544,48 @@ def noise_of_length(self, audio_in: Signal) -> Signal:
...
+    def _forward(self) -> Signal:
+        """
+        Creates a batch size number of LFO modulation signals with values [0,1]
+        """
+        rads = 2 * torch.pi * self.p("frequency").unsqueeze(1) / self.sample_rate
+        freqs = self.range.expand(self.batch_size, -1) * rads + self.phase.unsqueeze(1)
+        return ((1.0 + torch.cos(freqs)) / 2.0).as_subclass(Signal)
+
+
 class CrossfadeKnob(SynthModule):
```
maxsolomonhenry, line 589: Not sure how we want to handle this, but we'll want a modulation mixer at some point.

jorshi: Yes definitely -- I think this modulation routing and mixing can happen within Voices?

PR #195, opened 2021-03-20
Param freeze

jorshi:


torchsynth/synth.py
```diff
@@ -124,13 +129,33 @@ def randomize(self, seed: Optional[int] = None):
             # Profile to make sure this isn't too slow?
             mt19937_gen = csprng.create_mt19937_generator(seed)
             for parameter in self.parameters():
+                if isinstance(parameter, ModuleParameter) and parameter.frozen:
```
turian, line 132: Should we use this isinstance check for everything? Like not do `parameter.data.uniform_(0, 1, generator=mt19937_gen)` without it?

torchsynth/synth.py
```diff
@@ -79,13 +80,17 @@ def add_synth_modules(self, modules: List[Tuple[str, SynthModule]]):
...
-        Set parameters for synth by passing in a dictionary of modules and parameters
+        Set parameters for synth by passing in a dictionary of modules and parameters.
+        Can optionally freeze a parameter at this value to prevent further updates.
         """
         for (module_name, param_name), value in params.items():
             module = getattr(self, module_name)
             module.set_parameter(param_name, value.to(self.device))
+            # Freeze this parameter at this value now if freeze is True
+            if freeze:
+                module.get_parameter(param_name).frozen = True
```
turian, line 93: My proposal is that we add a convenience method

set_frozen_parameters where you just pass in a scalar and it turns it into the right batch size

That creates the tensor on self.device and calls self.set_parameters.

turian: Since in most cases you would probably want to freeze like that.

(Also, because then it give us a consistent naming scheme, like:

synth1B1-keyboard.midi_f0=69- 

even tho that's a little verbose

examples/examples.py
```diff
@@ -515,6 +515,32 @@ def stft_plot(signal, sample_rate=DEFAULT_SAMPLE_RATE):
         ipd.Audio(voice_out[i].cpu().detach().numpy(), rate=voice.sample_rate.item())
     )
 
+# Parameters can be set and frozen before randomization as well
+
+# +
+midi_f0 = torch.full((voice.batch_size.item(),), 42.0)
+duration = torch.full((voice.batch_size.item(),), 3.0)
+tuning = torch.full((voice.batch_size.item(),), 0.0)
```
turian, line 523: See comment below about `set_frozen_parameters` ?

turian (approved):
This generally looks good to me, but consider my small comments

examples/examples.py
```diff
@@ -515,6 +515,32 @@ def stft_plot(signal, sample_rate=DEFAULT_SAMPLE_RATE):
         ipd.Audio(voice_out[i].cpu().detach().numpy(), rate=voice.sample_rate.item())
     )
 
+# Parameters can be set and frozen before randomization as well
+
+# +
+midi_f0 = torch.full((voice.batch_size.item(),), 42.0)
+duration = torch.full((voice.batch_size.item(),), 3.0)
```
turian, line 522: Call this note_on_duration?

torchsynth/synth.py
jorshi: It's a good question -- is there a chance that we might have a non-ModuleParameter that we would want to randomize?

torchsynth/synth.py
jorshi: I like this -- will add.

torchsynth/synth.py
turian: Let's change it to this:

```
else:
   raise ValueError(f"{param} is not a ModuleParameter")
```

Then we'll know for sure.

torchsynth/synth.py
jorshi: Added just below

PR #196, opened 2021-03-20
LFO Modulation in Voice

jorshi:


torchsynth/module.py
```diff
@@ -586,6 +586,64 @@ def _forward(self) -> Signal:
         return ((1.0 + torch.cos(freqs)) / 2.0).as_subclass(Signal)
 
 
+class ModulationMixer(SynthModule):
```
jorshi, line 589: This is similar to the FaderBank mixer that I was working on in the FM branch, but modified to accept input signals and mix them together using the mix levels parameters.

torchsynth/synth.py
```diff
@@ -158,12 +166,28 @@ def _forward(self) -> T:
...
         # the same note_on_duration for both ADSRs.
         midi_f0, note_on_duration = self.keyboard()
-        pitch_envelope = self.pitch_adsr.forward(note_on_duration)
-        amp_envelope = self.amp_adsr.forward(note_on_duration)
 
-        audio_out = self.vco_1.forward(midi_f0, pitch_envelope)
-        vco_2_out = self.vco_2.forward(midi_f0, pitch_envelope)
+        # Compute LFOs with envelopes
+        lfo_1 = self.lfo_1() * self.lfo_1_adsr(note_on_duration)
+        lfo_2 = self.lfo_2() * self.lfo_2_adsr(note_on_duration)
```
jorshi, line 172: Open to different setups on this. two LFOs with different envelopes applied to them leads to some cool movement between the two modulation signals. And then these are mixed at various levels with the pitch and amplitude envelope.

torchsynth/module.py
```diff
@@ -393,16 +376,14 @@ def _forward(self, midi_f0: T, mod_signal: Signal) -> Signal:
...
 
         control_as_frequency = self.make_control_as_frequency(midi_f0, mod_signal)
         assert (control_as_frequency >= 0).all() and (
             control_as_frequency <= self.nyquist
         ).all()
 
         cosine_argument = self.make_argument(control_as_frequency)
-        cosine_argument += self.phase.unsqueeze(1)
-        self.phase.data = cosine_argument[:, -1]
+        cosine_argument += self.p("initial_phase").unsqueeze(1)
```
jorshi, line 386: Using initial phase here directly instead of the phase buffer. Because the phase buffer is initialized with the initial phase param during init, it doesn't get modified during randomization which means the phase of the oscillators is not deterministic. How do we feel about just using this here for now?

torchsynth/module.py
```diff
@@ -586,6 +548,124 @@ def _forward(self) -> Signal:
...
+        else:
+            curves = [1.0] * n_input
+
+        # Need to create the parameter ranges before calling super().__init
+        self.parameter_ranges = []
+        for i in range(n_input):
+            self.parameter_ranges.append(
+                ModuleParameterRange(
+                    -80.0,
+                    0.0,
```
jorshi, line 644: experimenting with making these mixing params in dB

torchsynth/module.py
```diff
@@ -586,6 +548,124 @@ def _forward(self) -> Signal:
...
+        params = torch.stack([self.p(p) for p in self.torchparameters], dim=1)
+        params = torch.pow(10.0, params / 20.0)
+
+        # Make sure we received the correct number of input signals
+        signals = torch.stack(signals, dim=1)
+        assert signals.shape[1] == params.shape[1]
+
+        # Mix signals and normalize output if required
+        output = torch.matmul(params.unsqueeze(1), signals).squeeze(1)
+        return util.normalize_if_clipping(output)
```
jorshi, line 666: do we just want to normalize all output? Without some sounds come out as essentially silence (you can't tell unless you turn off ipd.Audio normalization)

torchsynth/synth.py
```diff
@@ -125,12 +126,14 @@ def randomize(self, seed: Optional[int] = None):
...
                 parameter.data.uniform_(0, 1, generator=mt19937_gen)
-            for module in self._modules:
-                self._modules[module].seed = seed
         else:
             for parameter in self.parameters():
                 parameter.data = torch.rand_like(parameter, device=self.device)
 
+        # Add see to all modules
+        for module in self._modules:
+            self._modules[module].seed = seed
```
jorshi, line 135: Seed should always be set -- if randomize was previously called with a seed then that seed will remain in the modules unless we replace it with None when calling randomize without a seed

torchsynth/module.py
```diff
@@ -505,43 +486,24 @@ class VCA(SynthModule):
...
-            + "1 is all noise",
-        )
-    ]
-
-    def _forward(self, audio_in: Signal) -> Signal:
-        noise = self.noise_of_length(audio_in)
-        return util.crossfade2D(audio_in, noise, self.p("ratio"))
-
-    def noise_of_length(self, audio_in: Signal) -> Signal:
+    def _forward(self, device: Optional[torch.device] = None) -> Signal:
```
turian, line 498: Get rid of device and just use self.device

torchsynth/module.py
```diff
@@ -505,43 +486,24 @@ class VCA(SynthModule):
...
         if self.seed is not None:
             mt19937_gen = csprng.create_mt19937_generator(self.seed)
-            return audio_in.data.uniform_(-1, 1, generator=mt19937_gen)
+            noise = torch.empty((self.batch_size, self.buffer_size), device=device)
```
turian, line 501: device=self.device

torchsynth/module.py
```diff
@@ -505,43 +486,24 @@ class VCA(SynthModule):
...
+            noise.data.uniform_(-1, 1, generator=mt19937_gen)
         else:
-            return torch.rand_like(audio_in) * 2 - 1
+            noise = torch.rand((self.batch_size, self.buffer_size), device=device)
```
turian, line 504: Same

torchsynth/module.py
turian: @jorshi please create an issue for this, we will decide based upon MMD experiments

torchsynth/module.py
jorshi: device isn't an attribute of nn.Module, just the lightning stuff and SynthModule isn't a lightning module.

torchsynth/module.py
```diff
@@ -571,19 +533,180 @@ def __init__(
...
-        )
+
         # Create and save a buffer of sample points
         self.register_buffer("range", torch.arange(self.buffer_size))
 
+    def create_arg(self) -> Tuple[T, T]:
+        """
+        Creates the arg signal (phase over time)
+        """
+        frequency = self.p("frequency").unsqueeze(1)
```
maxsolomonhenry, line 544: Can you modulate the lfo frequency over time?

torchsynth/module.py
```diff
@@ -571,19 +533,180 @@ def __init__(
...
+
+    def _forward(self) -> Signal:
+        """
+        Creates a batch size number of LFO modulation signals with values [0,1]
+        """
+        arg, _ = self.create_arg()
+        return ((1.0 + torch.cos(arg)) / 2.0).as_subclass(Signal)
+
+
+class SquareSawLFO(AbstractLFO):
```
maxsolomonhenry, line 570: We could get away with a much simpler square/saw wave oscillator for LFO's because we're not worried about aliasing.

maxsolomonhenry: Could have a simple sine, square, saw, and maybe triangle. Mickey mouse though, nice and easy.

torchsynth/module.py
jorshi: Removed dB

torchsynth/module.py
jorshi: #202

PR #212, opened 2021-03-23
Hyperparameter getting and setting

turian:


tests/test_parameter.py
```diff
@@ -31,12 +31,15 @@ def test_construction(self):
         assert param_range.curve == 2.0
         assert param_range.curve == 2.0
 
+    # I think string unit tests are pointless
```
jorshi, line 34: Let's ditch it!

jorshi (approved):
Nice, looks clean -- let's get rid of that string test and examples.ipynb needs to be updated then it's good to go.

PR #214, opened 2021-03-23
Hyperparams idea

jorshi:
Cleaner way to get all the ModuleParameters for a synth keyed on the unique combination of module name and parameter name. Use this to set hyperparams.

### DIFF
--- a/torchsynth/synth.py
+++ b/torchsynth/synth.py
@@ -85,7 +85,7 @@ class AbstractSynth(LightningModule):
         self, include_frozen: Optional[bool] = False
     ) -> Dict[Tuple[str, str], ModuleParameter]:
         """
-        Property that returns a dictionary of ModuleParameters for this synth keyed
+        Returns a dictionary of ModuleParameters for this synth keyed
         on a tuple of the SynthModule name and the parameter name
         """
         parameters = []
--- a/torchsynth/synth.py
+++ b/torchsynth/synth.py
@@ -82,9 +82,21 @@ class AbstractSynth(LightningModule):
             self.add_module(name, module)
 
     @property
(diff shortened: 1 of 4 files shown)


torchsynth/synth.py
```diff
@@ -150,28 +176,29 @@ def test_step(self, batch, batch_idx):
...
-            if not ModuleParameter.is_parameter_frozen(parameter):
-                hparams.append(((name, "curve"), parameter.parameter_range.curve))
-                hparams.append(
-                    ((name, "symmetric"), parameter.parameter_range.symmetric)
-                )
+        for param_key, parameter in self.get_parameters().items():
+            hparams.append(((*param_key, "curve"), parameter.parameter_range.curve))
+            hparams.append(
+                ((*param_key, "symmetric"), parameter.parameter_range.symmetric)
+            )
```
turian, line 189: Wait a second, the signature changed to `Tuple[str, str, str]`. Where's the triple?

torchsynth/synth.py
```diff
@@ -81,6 +81,32 @@ def add_synth_modules(self, modules: List[Tuple[str, SynthModule]]):
 
             self.add_module(name, module)
 
+    def get_parameters(
+        self, include_frozen: Optional[bool] = False
```
turian, line 85: Shouldn't be Optional[bool], just bool

torchsynth/synth.py
turian: Maybe be more explicit here rather than tricky like:

```
for (module_name, parameter_name), parameter in self.get_parameters().items()
```

turian: We also just need the keys, not the full items()

PR #224, opened 2021-03-26
Started to kludge fp16 together

turian:
This actually isn't working for me, on CPU it errs out with:

```
  File "/Users/joseph/dev/torchsynth/torchsynth/module.py", line 388, in _forward
    assert (control_as_frequency >= 0).all() and (
AssertionError
```

A lot of ops aren't implemented for half on CPU (but maybe for GPU) so i just do casts to kludge it together.


turian:
@jorshi "so. fp16 appears to sorttt of work. But there are some issues. Lots of weird numerical errors leading to broken envelopes and aliased ass waveforms"

PR #225, opened 2021-03-26
All modules now know their device

turian:
I'm trying to make it easier for us to keep things on the right device.


examples/examples.py
```diff
@@ -170,8 +170,16 @@ def stft_plot(signal, sample_rate=DEFAULT_SAMPLE_RATE):
...
-    attack=a, decay=d, sustain=s, release=r, alpha=alpha, synthglobals=synthglobals
+    attack=a,
+    decay=d,
+    sustain=s,
+    release=r,
+    alpha=alpha,
+    synthglobals=synthglobals,
+    device=device,
 ).to(device)
+# I'm curious here if just doing device=device is enough and we
```
turian, line 181: @jorshi thoughts? I kinda want it set up so you never really need to do `.to` if you pass in the device.

torchsynth/globals.py
```diff
@@ -29,6 +30,11 @@ def __init__(
         self.sample_rate = sample_rate
         self.buffer_size = buffer_size
 
+    def to(self, device: torch.device):
+        self.batch_size = self.batch_size.to(device)
+        self.sample_rate = self.sample_rate.to(device)
+        self.buffer_size = self.buffer_size.to(device)
```
jorshi, line 36: Nice

jorshi (commented):
Okay this is really cool. But I think we can clean this up a bit. I'd like it if the device parameter in SynthModule had a default to None (this is what torch default to). Then we don't need to always pass it in. Also, I think we should override the `to` method of `nn.Module` in SynthModule something like this:

```
def to(self, device=None, **kwargs):
    super().to(device=device, **kwargs)
    self.synthglobals.to(device)
    self.device = device
```

And then in init, after we add all the parameters, we can call `super().to(device=device)` this will add all the parameters, synthglobals to the device as well as keep the device attribute in sync with the actual device.

torchsynth/globals.py
turian: Only thing I don't like is usually .to returns a copy and doesn't modify stuff in place

torchsynth/module.py
```diff
@@ -58,12 +65,16 @@ def __init__(self, synthglobals: SynthGlobals, **kwargs: Dict[str, T]):
...
+                # Put all the parameter ranges onto the device
+                parameter_range.to(self.device)
             self.add_parameters(
                 [
                     ModuleParameter(
                         value=None,
                         parameter_name=parameter_range.name,
-                        data=torch.rand((self.synthglobals.batch_size,)),
+                        # Should we use CSPRNG here? Or doesn't matter?
+                        data=torch.rand((self.synthglobals.batch_size,), device=device),
```
jorshi, line 77: I don't think it matters? We get the deterministic sampling from randomize(1) right? which randomizes all the parameters anyways. Don't think it matters during initialization.

PR #227, opened 2021-03-26
Remove filters

turian:


tests/test_module.py
```diff
@@ -16,6 +16,9 @@ class TestSynthModule:
     Tests for SynthModules
     """
 
+    # Disabling tests that relied upon deprecated synth
```
turian, line 19: This is cheesy but I want to get this out the door

PR #231, opened 2021-03-27
Device modifications

jorshi:
Some updates to make sure things are on the correct device. 

Things I learned while doing this -- having 0d tensors on the GPU does not necessarily lead to a speed up. I think creating a scalar tensor on the GPU is slower than just using a native Number type for most operations and comparisons. Especially if those tensors are different dtypes. Having parameter ranges on the GPU did not improve performance and actually resulted in worse performance (~ 15ms on batch size of 64) . Same with have the buffer size and batch size on the GPU. Having sample rate as a float did help a little bit though. Some of the assertions are slow. Especially the one I marked in the Parameters.


torchsynth/module.py
```diff
@@ -49,28 +58,38 @@ def __init__(self, synthglobals: SynthGlobals, **kwargs: Dict[str, T]):
...
+        self.synthglobals.to(device)
         self.torchparameters: nn.ParameterDict = nn.ParameterDict()
         # If this module needs a random seed, here it is
         self.seed: Optional[int] = None
 
         if self.parameter_ranges:
+            # TODO: Is this copying thing what we want?
+            # We want to create copies of the parameter ranges otherwise each instance
+            # of the same module type (ex. ADSR) will reference the same param range
+            parameter_ranges = copy.deepcopy(self.parameter_ranges)
```
jorshi, line 71: @turian -- not sure how you had the optuna hyperparam tuning going so you may have already thought about this, but basically every instance of an ADSR shared the same hyperparams, so if we updated the attack curve of one of the adsrs, then that would be the curve for all of them. If we want to adjust them all independently then we want this deepcopy approach here so each parameter has their own hyperparameters. Alternatively we decide that each instances of an ADSR should have the same hyperparams, in which case we maybe update our API a bit. However, now that I'm thinking about this maybe you had it setup correctly to begin with and I mucked it up with my hyperparam changes ... :/

torchsynth/module.py
```diff
@@ -98,7 +117,7 @@ def to_buffer_size(self, signal: Signal) -> Signal:
     def seconds_to_samples(self, seconds: T) -> T:
         # Do we want this?
         # assert seconds.ndim == 1
-        return torch.round(seconds * self.sample_rate).int()
+        return torch.round(seconds * self.sample_rate)
```
jorshi, line 120: Better to keep things as floats

turian:
Wow surprising insights to learn

examples/profiling_voice.py
```diff
@@ -0,0 +1,28 @@
...
+# %autoreload 2
+# -
+
+if torch.cuda.is_available():
+    device = 'cuda'
+else:
+    device = 'cpu'
+
+synthglobals64 = SynthGlobals(
+    batch_size=T(16), sample_rate=T(44100), buffer_size=T(4 * 44100)
```
turian, line 18: Just curious if we should profile:
a) what happens if the SynthGlobals tensors are on the device
b) Other keep them as ints for simplificity?

tests/test_module.py
```diff
@@ -74,6 +74,22 @@ def test_p(self):
         assert module.p("param_1") == 5000.0
     """
 
+    def test_set_parameter(self):
+        synthglobals = torchsynth.globals.SynthGlobals(batch_size=T(2))
+        module = synthmodule.ADSR(synthglobals, attack=T([0.5, 1.0]))
```
turian, line 79: Just another note here, if floats actually are faster, then maybe we should consider a *separate* PR where we convert all the scalars to plain floats and see the performance changes?

torchsynth/globals.py
```diff
@@ -31,9 +31,8 @@ def __init__(
         self.buffer_size = buffer_size
 
     def to(self, device: torch.device):
-        self.batch_size = self.batch_size.to(device)
-        self.sample_rate = self.sample_rate.to(device)
-        self.buffer_size = self.buffer_size.to(device)
+        # Only helpful to have sample rate on GPU, and as a float
```
turian, line 34: Interesting, any guesses why?

torchsynth/module.py
turian: hmmmm I wonder where this bug entered. How did you see it? Also, does this fix completely prevent anyone from footgunning themselves with this in the future?

My current optuna runs which are on main (2b416f98ab6756b672f2d9ec4cbc33c387540b97) don't appear to have tied ADSR params:

image

image

Like those look fine to me?

turian: so somehow this came thru main, or thru my PR, or thru yours. I think we should understand it a little more deeply just to make sure it doesn't shoot us in the foot.

turian: Okay shit this bug is in main.

turian: But your fix appears to work

torchsynth/module.py
turian: image

if this is the only time we use it, to create ramps, maybe we don't even want to round it.

turian: And just make a comment explaining that

torchsynth/module.py
```diff
@@ -193,6 +205,21 @@ def p(self, parameter_id: str) -> T:
...
+        """
+        Overriding the to call for nn.Module to transfer this module to device. This
+        makes sure that the ParameterRanges for each ModuleParamter and the globals
+        are also transferred to the correct device.
+        """
+        self.update_device(device)
+        return super().to(device=device, **kwargs)
+
+    def update_device(self, device=None):
+        self.synthglobals.to(device)
```
turian, line 218: Make a comment here explaining why we move parameter ranges to device but not the parameters themselves?

torchsynth/module.py
turian: What I don't love about this fix tho is that if you're in a module and you do self.parameter_ranges, you get the wrong thing.

What about either:

1) `self.parameter_ranges = copy.deepcopy(self.parameter_ranges)`
with a comment explaining that otherwise they are tied across module instances. Seems clean and works?

2) Rename it to orig_parameter_ranges in each module so people don't footgun themselves with it.

torchsynth/module.py
```diff
@@ -509,19 +536,19 @@ class Noise(SynthModule):
...
-    def _forward(self, device: Optional[torch.device] = None) -> Signal:
+    def _forward(self) -> Signal:
         if self.seed is not None:
             mt19937_gen = csprng.create_mt19937_generator(self.seed)
             self.noise.data.uniform_(-1, 1, generator=mt19937_gen)
         else:
-            self.noise = torch.rand((self.batch_size, self.buffer_size), device=device)
-            self.noise = self.noise * 2 - 1
+            self.noise.data.uniform_(-1, 1)
         return self.noise.as_subclass(Signal)
```
turian, line 552: Nice

torchsynth/module.py
```diff
@@ -721,7 +745,7 @@ def __init__(
                 )
             )
 
-        super().__init__(synthglobals=synthglobals, device=device, **kwargs)
+        super().__init__(synthglobals=synthglobals, **kwargs)
```
turian, line 748: okay nice pattern

turian: IMHO TBH then we can also remove SynthGlobals from each init since it's in kwargs

torchsynth/synth.py
```diff
@@ -226,6 +226,16 @@ def randomize(self, seed: Optional[int] = None):
...
 
+    def on_post_move_to_device(self) -> None:
+        """
+        LightningModule trigger after this Synth has been moved to a different device.
+        Use this to update children SynthModules device settings
+        """
+        self.synthglobals.to(self.device)
+        for module in self.modules():
+            if isinstance(module, SynthModule):
+                module.update_device(self.device)
```
turian, line 237: Okay cool but why not just module.to(self.device)

torchsynth/synth.py
```diff
@@ -238,30 +248,27 @@ def __init__(self, synthglobals: SynthGlobals, *args, **kwargs):
         # Register all modules as children
         self.add_synth_modules(
```
turian, line 249: We could clean up add_synth_modules so it plugs in synthglobals everywhere too.

So like you pass in:
```
["keyboard", MonophonicKeyboard]
["adsr_1", ADSR]
...
["mod_matrix", ModulationMixer, {"n_input": 4, "n_output": 5}]
```

But maybe the current way is cleanest and most explicit

But are you 100% sure that on_post_move_to_device is triggered in examples.py and the dumb non-lightning call:
```
voice = Voice().to(device)
```

I guess so? Since you can't create a voice on a device, you have to do `to`?

turian (approved):
Basically looks good to me. I have some questions and comments that we can address now or in a followup PR.

torchsynth/module.py
jorshi: Yes, good point. Unfortunately 1. doesn't work. I tried that approach. 2. is a good idea and would work. I'm into that.

As another point it would be great to be able to quickly save and load hyperparam settings. What if we had a json format for hyperparams for a particular synth? Say, Voice could have a default hyperparam setting for Synth1B1 that gets loaded automatically. But others could also be loaded as well.

torchsynth/module.py
jorshi: Yup, I like that idea

examples/profiling_voice.py
jorshi: ya I'll make an issue for the profiling stuff for each module -- Or a project perhaps and keep things organized for each module

tests/test_module.py
jorshi: Lets do it.

torchsynth/globals.py
jorshi: So the difference between it being a plain float on the CPU and a float on the GPU was pretty small. But the really slow down happened when it was an int on the GPU. I think that was the same for the other attributes -- my thought (needs to be tested) is that mixing different types of tensors on the GPU is what causes slowdowns.

torchsynth/module.py
jorshi: sure I dig that

torchsynth/synth.py
jorshi: soo this is another weird one. When you call `module.to(self.device)` on the module it goes through the process of moving all the parameters to the GPU -- BUT, those parameters have already been handled by this Synth because all parameters and buffers are known to it. So we really just want to update the stuff that we are managing. It is not only redundant to call it again, but it actually has a negative affect on performance during run. I'm not entirely clear on that part.

torchsynth/synth.py
jorshi: Yup it is definitely triggered in examples.py, if it wasn't then it wouldn't work (we'd get different device complaints). I like your clean up idea though. I think that is great since then we don't need to pass in synthglobals for each one.

torchsynth/module.py
jorshi: Issue for saving and loading hyperparam settings: #232

examples/profiling_voice.py
jorshi: Made a project with issues for profiling individual modules

tests/test_module.py
jorshi: #239

torchsynth/module.py
jorshi: pulling back on this one. synthglobals is a required positional argument in SynthModule so it's actually not in kwargs in that init.

torchsynth/synth.py
```diff
@@ -55,30 +55,32 @@ def buffer_size(self) -> T:
...
 
-            if module.buffer_size != self.buffer_size:
-                raise ValueError(f"{module} buffer size does not match")
+            if not issubclass(module, SynthModule):
+                raise TypeError(f"{module} is not a SynthModule")
 
-            self.add_module(name, module)
+            self.add_module(
+                name, module(self.synthglobals, device=self.device, **params)
+            )
```
jorshi, line 83: This is a cool pattern, good idea @turian  -- now if any modules are added to the Synth after construction then they constructed on the device. We don't have anything like that now, but could allow for some interesting modular architectures

jorshi: Also nice b/c we don't have to have all those checks on synthglobals

examples/profiling_voice.py
```diff
@@ -0,0 +1,26 @@
...
+import torch
+import torch.tensor as T
+
+from torchsynth.globals import SynthGlobals
+from torchsynth.synth import Voice
+
+# %load_ext autoreload
+# %autoreload 2
+# -
+
```
turian, line 11: Can we remove this until it's a little more mature?

torchsynth/module.py
turian: Right but just because it's a required kwarg, if a submodule has it in kwargs should be fine no?

torchsynth/synth.py
turian: Really has a negative impact? I thought stuff already on the device is a no-op.
If so can you just add this comment?

torchsynth/module.py
jorshi: ahhh ya I understand, yes we can get rid of the synthglobals from the super init.

torchsynth/synth.py
jorshi: Will make a comment and an issue to check out. I'm not sure what's going on there -- might be the override on to in SynthModule?

examples/profiling_voice.py
jorshi: Oh ya totally

PR #245, opened 2021-03-29
Module autogen style updated

Doddox-UV:
Modules generated in separate pages...


docs/source/filter.md
```diff
@@ -0,0 +1,9 @@
+Torchsynth.filter
```
jorshi, line 1: We've removed filters so you can get rid of this file

docs/source/index.md
```diff
@@ -56,9 +56,14 @@ contributing/contributing
 ---
 maxdepth: 3
 name: modules
-caption: Modules
+caption: Torchsynth Packages
 ---
-modules
+defaults
+filter
```
jorshi, line 62: Same here, remove filters

requirements.txt
```diff
@@ -23,7 +23,8 @@ pygments>=2.7.4 # not directly required, pinned by Snyk to avoid a vulnerability
 https://github.com/PyTorchLightning/lightning_sphinx_theme/tarball/master#egg=pt-lightning-sphinx-theme
 sphinxcontrib-napoleon
 mock
-recommonmark
+myst_parser
+linkify-it-py
```
jorshi, line 27: I'm getting an error that sphinx_rtd_theme package is also required when trying to build. Does that make sense? If so can you add in here?

jorshi: And in setup.py too

jorshi (changes requested):
Looks good! A couple comments, afterwards those are addressed this is good to merge.

requirements.txt
Doddox-UV: Yes it makes sense because "sphinx_rtd_theme" is called by the conf.py file

Doddox-UV: So I add it in the configuration: setup.py and requirements.txt

docs/source/index.md
Doddox-UV: So I can remove filter.md from the project. Right?

jorshi: Yep, go ahead, thank you

Doddox-UV: Done

PR #246, opened 2021-03-30
Control rate

jorshi:
Update that adds control rate SynthModules with optional upsampling. Includes some performance updates for ADSR as well.


torchsynth/module.py
```diff
@@ -227,7 +227,65 @@ def update_device(self, device=None):
...
+                self.synthglobals.buffer_size, mode="linear", align_corners=True
+            )
+
+    @property
+    def sample_rate(self) -> T:
+        """
+        The internal sample rate for this module is control rate
+        """
+        assert self.synthglobals.control_rate.ndim == 0
+        return self.synthglobals.control_rate
```
jorshi, line 258: Would like your guys opinion on this idea for overriding sample rate and buffer size properties to return the control rate and control buffer size. The benefit of this is that within the module the code will just run nicely at the correct rate and any calls to helper functions like seconds_to_samples will work without having to think about the different rates. The downside is that it is maybe a bit too cute and could lead to confusion, for example if you run this module with upsampling then from an outside perspective the sample_rate property won't seem right. But internally it would be.

torchsynth/synth.py
```diff
@@ -252,12 +252,12 @@ def __init__(self, synthglobals: SynthGlobals, *args, **kwargs):
...
-                ("lfo_1", LFO),
-                ("lfo_2", LFO),
-                ("lfo_1_amp_adsr", ADSR),
-                ("lfo_2_amp_adsr", ADSR),
+                ("adsr_1", ADSR, {"upsample": True}),
+                ("adsr_2", ADSR, {"upsample": True}),
+                ("lfo_1", LFO, {"upsample": True}),
+                ("lfo_2", LFO, {"upsample": True}),
+                ("lfo_1_amp_adsr", ADSR, {"upsample": True}),
+                ("lfo_2_amp_adsr", ADSR, {"upsample": True}),
```
jorshi, line 260: Now in the synth we just need the modulation modules that output to audio rate modules to upsample their output. Other modules like the adsrs for the LFOs output at control rate.

examples/examples.py
```diff
@@ -199,11 +199,18 @@ def stft_plot(signal, sample_rate=DEFAULT_SAMPLE_RATE):
...
+# **Generating Random Envelopes**
+#
+# If we don't set parameters for an ADSR, then the parameters will be random when
+# initialized. We can also pass in an upsample attribute to allow control signal
+# modules to upsample to full sample rate.
+
 # Note that module parameters are optional. If they are not provided,
 # they will be randomly initialized (like a typical neural network module)
-adsr = ADSR(synthglobals, device)
+adsr = ADSR(synthglobals, device=device, upsample=True)
```
turian, line 210: Just a quick question, why would we want upsample=False?

tests/test_globals.py
```diff
@@ -0,0 +1,29 @@
...
+from torchsynth.globals import SynthGlobals
+from torchsynth.default import DEFAULT_SAMPLE_RATE, DEFAULT_BUFFER_SIZE
+
+
+def test_synth_globals():
+    synthglobals = SynthGlobals(T(64))
+    assert synthglobals.sample_rate == DEFAULT_SAMPLE_RATE
+    assert synthglobals.batch_size == 64
+    assert synthglobals.buffer_size == DEFAULT_BUFFER_SIZE
+    assert synthglobals.control_rate == DEFAULT_SAMPLE_RATE // 100
```
turian, line 19: Why is 100 a magic number here?

turian: Should we make it a default or at least name it?

torchsynth/globals.py
```diff
@@ -30,12 +34,30 @@ def __init__(
         self.sample_rate = sample_rate
         self.buffer_size = buffer_size
 
+        # Default control rate to 100 times less than audio rate
+        if control_rate is None:
+            self.control_rate = (self.sample_rate // 100).clone().detach()
```
turian, line 39: 1) Let's not use magic numbers, make this something in defaults?
2) If sample_rate % control == 0 maybe we should throw an assert? Because then julian resampling will be expensive?

torchsynth/globals.py
```diff
@@ -30,12 +34,30 @@ def __init__(
...
+        else:
+            assert control_rate.ndim == 0
+            self.control_rate = control_rate
+
+        # Buffer size for control signals
+        self.control_buffer_size = (
+            torch.ceil((buffer_size / sample_rate * self.control_rate))
```
turian, line 46: If the above conditions hold then we assert control_rate * buffer_size // sample_rate = this ceil value?

torchsynth/module.py
```diff
@@ -227,7 +227,65 @@ def update_device(self, device=None):
             parameter.parameter_range.to(device)
 
 
-class ADSR(SynthModule):
+class ControlRateModule(SynthModule):
```
turian, line 230: Now you get me excited to add ParameterModule and Signal Module :)

torchsynth/module.py
```diff
@@ -227,7 +227,65 @@ def update_device(self, device=None):
...
+    at a control rate as opposed to full audio rate.
+
+    Optionally, the output control signal can be upsampled to the correct
+    sample rate and buffer size.
+    """
+
+    def __init__(
+        self,
+        synthglobals: SynthGlobals,
+        upsample: Optional[bool] = False,
```
turian, line 242: Can you document when upsample should and should not be true? with maybe an example when you wouldn't want to?

torchsynth/module.py
turian: Hmm how hard would be only to have sample_rate for SignalModules and control_rate for control_modules?

I would assume sample_rate of a ControlRateModule would including the upsampling

torchsynth/module.py
```diff
@@ -227,7 +227,65 @@ def update_device(self, device=None):
...
+        assert self.synthglobals.control_buffer_size.ndim == 0
+        return self.synthglobals.control_buffer_size
+
+    def forward(self, *args: Any, **kwargs: Any) -> Signal:  # pragma: no cover
+        """
+        Wrapper for _forward that ensures a buffer_size length output.
+        """
+        signal = self._forward(*args, **kwargs)
+        if self.upsample:
+            signal = self.upsample(signal.unsqueeze(1)).squeeze(1)
```
turian, line 274: Again, I still think control rate should factor nicely into the signal_rate so we don't need to fix lengths.

torchsynth/module.py
```diff
@@ -284,11 +356,10 @@ def _forward(self, note_on_duration: T) -> Signal:
         # Calculations to accommodate attack/decay phase cut by note duration.
         attack = self.p("attack")
         decay = self.p("decay")
+        self.alpha = self.p("alpha").unsqueeze(1)
```
turian, line 359: Why do we need unsqueeze here?

torchsynth/module.py
```diff
@@ -284,11 +356,10 @@ def _forward(self, note_on_duration: T) -> Signal:
...
 
         new_attack = torch.minimum(attack, note_on_duration)
-        new_decay = torch.maximum(
-            note_on_duration - attack, T([0.0], device=self.device)
-        )
+        new_decay = torch.maximum(note_on_duration - attack, self.zero)
```
turian, line 362: Nice

torchsynth/module.py
```diff
@@ -613,11 +689,13 @@ def _forward(self, mod_signal: Signal) -> Signal:
         """
         Must be implemented in deriving classes
         """
+        # This module accepts signals at control rate!
```
turian, line 692: hmmm yeah okay maybe SignalModule is the wrong word, since we really mean ControlRateModule vs AudioModule or something

torchsynth/synth.py
turian: Can we be explicit about when upsampling does or doesn't happen and make it required?

The default scares me a little

examples/examples.py
jorshi: I think you saw below -- but when we modulate another control rate module like LFO we need the output to be control rate.

torchsynth/globals.py
jorshi: Not using Julian resampling so I don't think it will matter too much if it isn't divisible -- but, I think it's not a bad idea anyway to have it as a default and have an assert. Makes more sense for it to be divisible anyway. Agree on magic numbers

torchsynth/globals.py
jorshi: Yes sounds good

torchsynth/module.py
jorshi: Right !? Like your comment below I like the idea of AudioRateModule and ParameterModule -- could be cool. It's actually giving me some ideas now ...

torchsynth/module.py
jorshi: Yes. Also like the idea of it not being a default

torchsynth/module.py
jorshi: Not hard. ControlRateModule derive from SynthModule currently and has access to everything.

torchsynth/module.py
jorshi: Agreed. This fix length is just to be consistent with SynthModule since that runs fix_length on every output

torchsynth/module.py
jorshi: We always need to unsqueeze parameters before applying them to signals. parameters are 1-dimensional and signals are 2-dimensional. Sometimes we just need to do stuff with the param values before unsqueezing, which is why attack and decay don't have that above -- they just get it later.

torchsynth/synth.py
jorshi: Yes let's drop the default.

torchsynth/globals.py
turian: What are you using if not Julian?

examples/examples.py
turian: Got it yeah

torchsynth/globals.py
jorshi: PyTorch Upsample with linear interpolation

torchsynth/module.py
turian: Right but I thought squeeze and unsqueeze makes a copy. We could probably get more performance by changing squeeze and unsqueeze to views: https://github.com/turian/torchsynth/issues/249

For another PR :)

torchsynth/globals.py
jorshi: hmm, so this gets weird with some buffer sizes -- I don't think we can make this assertion. Ex: say we have a buffer size of 4096, sr of 48k, and a control rate of 480.  `ceil(480 * 4096 /48000) = 41`, `control_rate * buffer_size // sample_rate = 40`

I think this will be fine to leave as is. The upsample operation works by upsampling to buffer size number of samples, so the number of samples that goes in doesn't really matter. If anything actually we should round and not ceil so we get the closest number of samples.

turian (approved):

PR #247, opened 2021-03-30
Noise performance

jorshi:


torchsynth/config.py
```diff
@@ -0,0 +1,37 @@
+import warnings
+import torch
+
```
jorshi, line 3: Wasn't sure about the name of this module. It is sort of module config type stuff. This function will run when someone imports torchsynth, so they will know right away whether they can produce reproducible results.

torchsynth/synth.py
```diff
@@ -214,15 +214,17 @@ def randomize(self, seed: Optional[int] = None):
...
         if seed is not None:
-            # Profile to make sure this isn't too slow?
-            mt19937_gen = csprng.create_mt19937_generator(seed)
+            torch.manual_seed(seed)
             for parameter in self.parameters():
                 if not ModuleParameter.is_parameter_frozen(parameter):
-                    parameter.data.uniform_(0, 1, generator=mt19937_gen)
+                    new_params = torch.rand(
+                        (self.batch_size,), device="cpu", pin_memory=True
+                    )
```
jorshi, line 222: torch rand was faster by 150ms over 64 iterations at batch size of 64. It's small but adds up to ~10hrs on Synth1B1

examples/examples.py
```diff
@@ -410,7 +410,7 @@ def stft_plot(signal, sample_rate=DEFAULT_SAMPLE_RATE):
 keyboard = MonophonicKeyboard(synthglobals, device=device)
 sine = SineVCO(synthglobals, device=device)
 square_saw = SquareSawVCO(synthglobals, device=device)
-noise = Noise(synthglobals, device=device)
+noise = Noise(synthglobals, 123, device=device)
```
turian, line 413: seed=123

tests/test_synth.py
```diff
@@ -92,6 +92,18 @@ def test_deterministic_noise(self):
         assert torch.mean(torch.abs(x11 - x2)) > 1e-6
         assert torch.mean(torch.abs(x11 - x12)) < 1e-6
 
+        if self.device == "cuda":
```
turian, line 95: Give a comment or better names to the synths so this test is easier to understand?

torchsynth/config.py
turian: reproducibility.py?

torchsynth/config.py
```diff
@@ -0,0 +1,37 @@
...
+            [
+                4.900934100151062012e-01,
+                8.964447379112243652e-01,
+                4.556279778480529785e-01,
+            ],
+        ]
+    )
+
+    torch.random.manual_seed(0)
+    sample = torch.rand((3, 3), dtype=torch.float)
```
turian, line 31: Specifiy CPU

torchsynth/config.py
```diff
@@ -0,0 +1,37 @@
+import torch
+
+
+def check_for_determinism():
+    """
+    This checks to make sure that the expected random results
+    are produced by torch.rand when seeded. This raises a warning
+    indicating if reproducibility is not guaranteed.
```
turian, line 8: Add a comment with the following details:
1) CPU vs GPU torch.rand gives different results (https://discuss.pytorch.org/t/deterministic-prng-across-cpu-cuda/116275)
2) Profiling indicated we could do torch.rand on CPU, faster than torchcsprng

torchsynth/config.py
```diff
@@ -0,0 +1,37 @@
...
+                4.556279778480529785e-01,
+            ],
+        ]
+    )
+
+    torch.random.manual_seed(0)
+    sample = torch.rand((3, 3), dtype=torch.float)
+    if not torch.all(sample.eq(expected)):
+        # TODO Make this a warning before we release v1
+        raise EnvironmentError(
```
turian, line 34: # pragma: no-cover

to avoid code cov complaints

torchsynth/config.py
```diff
@@ -0,0 +1,37 @@
...
+            "Random number generator produced incorrect results. "
+            "Reproducible dataset generation is not supported on this system."
```
turian, line 36: "Please file an issue on github [our issue link] with details about your CPU architecture and what random results you get"
[include their random output in the trace]

torchsynth/module.py
```diff
@@ -541,22 +541,23 @@ def _forward(self, audio_in: Signal, control_in: Signal) -> Signal:
 
 class Noise(SynthModule):
     """
-    Generates white noise that is the same length as the buffer
+    Generates white noise that is the same length as the buffer.
+
+    A batch size number of noise samples are pre-computed for performance.
```
turian, line 546: Hmm okay :\ I still would prefer a prime so that things cycle a bit

torchsynth/synth.py
```diff
@@ -214,15 +213,20 @@ def randomize(self, seed: Optional[int] = None):
         Randomize all parameters
         """
         if seed is not None:
-            # Profile to make sure this isn't too slow?
-            mt19937_gen = csprng.create_mt19937_generator(seed)
+            torch.manual_seed(seed)
             for parameter in self.parameters():
                 if not ModuleParameter.is_parameter_frozen(parameter):
-                    parameter.data.uniform_(0, 1, generator=mt19937_gen)
```
turian, line 221: Seems like we can remove the dependency on torch csprng entirely then?

torchsynth/synth.py
```diff
@@ -1,7 +1,6 @@
 from typing import Any, Dict, List, Optional, Tuple
 
 import torch
-import torchcsprng as csprng
 from pytorch_lightning.core.lightning import LightningModule
 from torch import tensor as T
```
turian, line 5: import reproducibility? Or that's an option for abstract synth?

jorshi: See above -- it's in the package __init__

torchsynth/module.py
jorshi: I messed around with this a bit. I found a clean and fast way to do it -- but things get weird as soon as we introduce the deterministic batch retrieval aspect. When we call synth(batch_idx) we need to make sure that the correct cycle of the noise is returned. Would need to spend a bit more time to think that through and implement in a way that doesn't slow things down too much. I'm still pretty convinced that is won't be a perceptually relevant thing. But, let's file an issue for it if you're okay with that.

torchsynth/synth.py
jorshi: Removed

tests/test_synth.py
```diff
@@ -92,6 +94,20 @@ def test_deterministic_noise(self):
         assert torch.mean(torch.abs(x11 - x2)) > 1e-6
         assert torch.mean(torch.abs(x11 - x12)) < 1e-6
 
+        # If a GPU is available then make sure the same
+        # tests as above are also deterministic on the GPU.
+        if self.device == "cuda":
+            synth.to("cuda")
+            synth.randomize(1)
+            cuda11 = synth()
+            synth.randomize()
```
turian, line 103: Can you pass an explicit seed here?

tests/test_synth.py
```diff
@@ -92,6 +94,20 @@ def test_deterministic_noise(self):
         assert torch.mean(torch.abs(x11 - x2)) > 1e-6
```
turian, line 94: Above can you call cpusynth

tests/test_synth.py
```diff
@@ -92,6 +94,20 @@ def test_deterministic_noise(self):
...
+        # If a GPU is available then make sure the same
+        # tests as above are also deterministic on the GPU.
+        if self.device == "cuda":
+            synth.to("cuda")
+            synth.randomize(1)
+            cuda11 = synth()
+            synth.randomize()
+            cuda2 = synth()
+            synth.randomize(1)
+            cuda12 = synth()
```
turian, line 106: Can you call these cudasynth?

tests/test_synth.py
```diff
@@ -92,6 +94,20 @@ def test_deterministic_noise(self):
...
+
+            assert torch.mean(torch.abs(cuda11 - cuda2)) > 1e-6
+            assert torch.mean(torch.abs(cuda11 - cuda12)) < 1e-6
```
turian, line 109: And then assert that cpusynth == cudasynth when we would expect that?

torchsynth/module.py
turian: I think this is okay for now.

torchsynth/reproducibility.py
```diff
@@ -0,0 +1,48 @@
...
+    to make sure that the expected random results are produced by torch.rand when
+    seeded. This raises an error indicating if reproducibility is not guaranteed.
+
+    Running torch.rand on CPU and GPU give different results, so all seeded
+    randomization where determinism is important occurs on the CPU and then is
+    transferred over to the GPU, if one is being used.
+    See https://discuss.pytorch.org/t/deterministic-prng-across-cpu-cuda/116275
+
+    torchcsprng allowed for determinism between the CPU and GPU, however
+    profiling indicated that torch.rand on CPU was more efficient.
```
turian, line 16: Could you also link this issue: https://github.com/pytorch/csprng/issues/126

torchsynth/synth.py
turian: Nice

torchsynth/module.py
turian: Actually we do need one fix. Create DEFAULT_BATCH_SIZE = 64 and always use that.

We want that if someone is doing batch_size 1024 for speed, they get *identical* results. Ya feel me?

turian: I meant, always use DEFAULT_BATCH_SIZE for white noise. And explain it's for reproducibility

turian: Might even consider making a test that 2 x batchsize=64 == 1 x batchsize 128 :)

torchsynth/module.py
```diff
@@ -541,22 +540,22 @@ def _forward(self, audio_in: Signal, control_in: Signal) -> Signal:
...
-    def __init__(self, synthglobals: SynthGlobals, **kwargs):
+    def __init__(self, synthglobals: SynthGlobals, seed: int, **kwargs):
         super().__init__(synthglobals, **kwargs)
-        self.register_buffer(
-            "noise",
-            torch.empty((self.batch_size, self.buffer_size), device=self.device),
-        )
+
+        # Create a prime number of noise samples, we will cycle through these
+        torch.random.manual_seed(seed)
```
turian, line 554: Wait no, we can't do this. There two pernicious anti-patterns in RNGs that people fall into:
1) Not using deterministic seeds when they should. It makes it harder to debug stuff retroactively. We've fixed this.
2) Changing the global seed in an inner loop.

Let's say someone calling our library is using torch.seed to randomize hyperparameters. BOOM now they are getting the same hyperparameters because we mucked with the global seed.

I would comb the code for any time we do random seeding (in random.seed, np.random.seed, pytorch_lightning.seed_everything, and torch.manual_seed) to make sure we don't mess with downstream clients of our code.

Here you probably want to do this: https://pytorch.org/docs/stable/random.html#torch.random.fork_rng

I can't really figure out how it works so actually....

It might be easier and safer to do this:

oldrngstate = torch.random.get_rng_state()
seed, do some random stuff, etc.
torch.random.set_rng_state(oldrngstate)

torchsynth/reproducibility.py
```diff
@@ -0,0 +1,49 @@
...
+            ],
+            [
+                4.900934100151062012e-01,
+                8.964447379112243652e-01,
+                4.556279778480529785e-01,
+            ],
+        ]
+    )
+
+    torch.random.manual_seed(0)
```
turian, line 39: Here again. We can't muck with the global RNG state. I'd do a grep for seed everywhere in the code, to make sure we fix this.

torchsynth/module.py
jorshi: Yes totally get this. I think we need to consider how parameters are going to be sampled then too, b/c those need to be updated as well to allow for larger batch sizes

torchsynth/module.py
jorshi: Good catch. I think we can avoid the global random number generator altogether and just create new ones that we can seed for these sort of operations.

torchsynth/module.py
```diff
@@ -541,23 +540,43 @@ def _forward(self, audio_in: Signal, control_in: Signal) -> Signal:
...
-        self.register_buffer(
-            "noise",
-            torch.empty((self.batch_size, self.buffer_size), device=self.device),
-        )
+
+        # Pre-compute noise
+        generator = torch.Generator(device="cpu").manual_seed(seed)
+        noise = torch.empty((DEFAULT_BATCH_SIZE, self.buffer_size), device="cpu")
+        noise.data.uniform_(-1.0, 1.0, generator=generator)
+        if self.batch_size > DEFAULT_BATCH_SIZE:
```
turian, line 557: hrmmmm why? can't we just do modulus

torchsynth/util.py
```diff
@@ -52,6 +52,29 @@ def normalize(signal: Signal) -> Signal:
     return signal / max_sample
 
 
+def batch_resize(x: torch.Tensor, size: int):
```
turian, line 55: So are we really sure we need this? Can we just insist that batch size is divisible by DEFAULT_BATCH_SIZE and be done

turian: like if we just assert batch_size % DEFAULT_BATCH_SIZE = 0 doesn't that make our lives easier?

torchsynth/synth.py
```diff
@@ -214,15 +214,25 @@ def randomize(self, seed: Optional[int] = None):
         Randomize all parameters
         """
         if seed is not None:
-            # Profile to make sure this isn't too slow?
-            mt19937_gen = csprng.create_mt19937_generator(seed)
+            self.generator.manual_seed(seed)
```
turian, line 217: If we only use self.generator here, then maybe let's just put it here? Will it be a performance hit? Otherwise it's clearer

torchsynth/synth.py
```diff
@@ -214,15 +214,25 @@ def randomize(self, seed: Optional[int] = None):
...
             for parameter in self.parameters():
                 if not ModuleParameter.is_parameter_frozen(parameter):
-                    parameter.data.uniform_(0, 1, generator=mt19937_gen)
+                    # TODO figure out how to support different batch sizes and
+                    # maintain reproducibility.
```
turian, line 221: Okay yeah a little thorny. So we can close this and move on and fix it later, make an issue, put it in the comments, and call it a day

torchsynth/synth.py
```diff
@@ -214,15 +214,25 @@ def randomize(self, seed: Optional[int] = None):
...
+                    if self.device.type != "cpu":  # pragma: no cover
```
turian, line 222: A little cleaner to be like self.device.type != self.generator.type OR just make self.generator a local variable and call it cpu_rng and be done with it

torchsynth/synth.py
turian: I have some ideas how to do it but let's just make an issue and put it in the code and do it later

torchsynth/module.py
```diff
@@ -541,23 +540,43 @@ def _forward(self, audio_in: Signal, control_in: Signal) -> Signal:
...
+        self.register_buffer("noise", noise.to(self.device))
+        self.offset = 0
 
     def _forward(self) -> Signal:
-        if self.seed is not None:
-            mt19937_gen = csprng.create_mt19937_generator(self.seed)
-            self.noise.data.uniform_(-1, 1, generator=mt19937_gen)
+        # Return noise quickly if we are returning the whole batch
+        # and don't need to offset at all.
+        if self.batch_size >= DEFAULT_BATCH_SIZE and self.offset == 0:
```
turian, line 566: okay this self.offset thing is a bit gnarly, but I think we should just revisit this.

I have a few ideas how to fix this later

torchsynth/synth.py
jorshi: negligible in terms of performance to switch out to here

PR #258, opened 2021-04-02
Profiling script

jorshi:
Not sure if we want to include this, but this has been mega helpful for me in profiling. Also used this with line profiler to look at line-by-line profiles https://github.com/pyutils/line_profiler


turian:
I'm totally fine with this, but let's call it torchsynth-profile.py, not torchsynth.profile

turian (approved):
Let's rename it to .py and then merge it

jorshi:
This is a pretty typical to leave out the .py for command line python scripts in packages. https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html -- with this you can call `torchsynth.profile` from the command line which is a bit more typical for a command line tool. 

If you're not crazy about having that file without the extension we could use a console-script entry point and move the profiling script into the main torchsynth source folder. https://python-packaging.readthedocs.io/en/latest/command-line-scripts.html#the-console-scripts-entry-point -- I would also be okay to rename to `torchsynth-profile` if you like that better.

PR #267, opened 2021-04-09
SynthGlobals => SynthConfig

turian:
Before getting fancy with the synthconfig, this is a mechanical rename


jorshi (approved):
Looks good

PR #269, opened 2021-04-09
Remove all defaults

turian:
Review after https://github.com/turian/torchsynth/pull/268


torchsynth/module.py
```diff
@@ -618,31 +620,34 @@ class Noise(SynthModule):
     these in a synth then choose different seeds for each instance.
     """
 
-    def __init__(self, synthglobals: SynthGlobals, seed: int, **kwargs):
-        super().__init__(synthglobals, **kwargs)
+    noise_batch_size: Final[int] = 64
```
turian, line 623: Just learned about this type, it means constant

torchsynth/module.py
```diff
@@ -618,31 +620,34 @@ class Noise(SynthModule):
...
-            noise = noise.repeat(self.batch_size // DEFAULT_BATCH_SIZE, 1)
+            noise = noise.repeat(self.batch_size // self.noise_batch_size, 1)
 
         self.register_buffer("noise", noise.to(self.device))
         self.offset = 0
 
     def _forward(self) -> Signal:
         # Return noise quickly if we are returning the whole batch
         # and don't need to offset at all.
-        if self.batch_size >= DEFAULT_BATCH_SIZE and self.offset == 0:
```
turian, line 645: @jorshi is this actually what we want?

jorshi: Yah I think so - in init batch_size number of random noise samples are created. If batch_size is greater than the noise_batch_size (64) then it repeats the noise samples. So if the offset is zero and the batch_size is greater than or equal to the noise batch size then we can just return the buffer of noise.

jorshi: I admit that it is confusing. Would be good to clean this up and simplify

pytest.ini
```diff
@@ -0,0 +1,3 @@
+[pytest]
+env = 
+    TORCHSYNTH_DEBUG=True
```
jorshi, line 3: cool

PR #283, opened 2021-04-09
Reproducibility

turian:
This is the easiest way for 1.0 to get reproducibility. We can fix after

### DIFF
--- a/torchsynth/config.py
+++ b/torchsynth/config.py
@@ -91,7 +91,7 @@ def check_for_reproducibility():
     """
     Reproducible results are important to torchsynth and Synth1B1, so we are testing
     to make sure that the expected random results are produced by torch.rand when
-    seeded. This raises an error indicating if reproducibilty is not guaranteed.
+    seeded. This raises an error indicating if reproducibility is not guaranteed.
 
     Running torch.rand on CPU and GPU give different results, so all seeded
     randomization where determinism is important occurs on the CPU and then is
--- a/torchsynth/config.py
+++ b/torchsynth/config.py
@@ -49,9 +49,15 @@ class SynthConfig:
         self.reproducible = reproducible
         if self.reproducible:
             # Currently, noise module (https://github.com/turian/torchsynth/issues/255)
-            # and abstract synth parameter randomization (https://github.com/turian/torchsynth/issues/253)
+            # and abstract synth parameter randomization
+            # (https://github.com/turian/torchsynth/issues/253)
             # are non-deterministic unless batch_size == 64.
-            assert batch_size == 64
+            if batch_size != 64:
+                raise ValueError(
+                    "Reproducibility currently only supported "
+                    "with batch_size = 64. If you want a different batch_size, "
+                    "initialize SynthConfig with reproducible=False"
+                )
             check_for_reproducibility()
 
         self.debug = debug
--- a/torchsynth/config.py
+++ b/torchsynth/config.py
@@ -45,6 +45,10 @@ class SynthConfig:
         self.control_rate = torch.tensor(control_rate)
         self.reproducible = reproducible
         if self.reproducible:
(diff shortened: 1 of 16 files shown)


torchsynth/config.py
```diff
@@ -78,7 +91,7 @@ def check_for_reproducibility():
     """
     Reproducible results are important to torchsynth and Synth1B1, so we are testing
     to make sure that the expected random results are produced by torch.rand when
-    seeded. This raises an error indicating if reproducibility is not guaranteed.
+    seeded. This raises an error indicating if reproducibilty is not guaranteed.
```
jorshi, line 94: Original spelling was correct

torchsynth/module.py
```diff
@@ -620,13 +620,18 @@ class Noise(SynthModule):
...
     noise_batch_size: int = 64
     # Unfortunately, Final is not supported until Python 3.8
     # noise_batch_size: Final[int] = 64
 
     def __init__(self, synthconfig: SynthConfig, seed: int, **kwargs):
         super().__init__(synthconfig, **kwargs)
 
+        # https://github.com/turian/torchsynth/issues/255
+        assert self.batch_size == 64 or not self.synthconfig.reproducible
+
```
jorshi, line 634: Let's dump all the offset stuff and just return the noise now that we are enforcing a batch size of 64 when reproducible

torchsynth/synth.py
```diff
@@ -36,9 +36,13 @@ class AbstractSynth(LightningModule):
...
 
-    def __init__(self, synthconfig: SynthConfig, *args, **kwargs):
+    def __init__(self, synthconfig: Optional[SynthConfig] = None, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.synthconfig = synthconfig
+        if synthconfig:
+            self.synthconfig = synthconfig
+        else:
+            # Use the default
+            self.synthconfig = SynthConfig()
```
jorshi, line 45: nice

torchsynth/synth.py
```diff
@@ -36,9 +36,13 @@ class AbstractSynth(LightningModule):
         buffer_size (int): number of samples expected at output of child modules
     """
 
-    def __init__(self, synthconfig: SynthConfig, *args, **kwargs):
+    def __init__(self, synthconfig: Optional[SynthConfig] = None, *args, **kwargs):
         super().__init__(*args, **kwargs)
-        self.synthconfig = synthconfig
+        if synthconfig:
```
jorshi, line 41: if synthconfig is not None

torchsynth/synth.py
```diff
@@ -172,6 +176,8 @@ def forward(
         """
         if batch_idx is not None:
             self.randomize(seed=batch_idx)
+        else:
+            assert not self.synthconfig.reproducible
```
jorshi, line 180: If we are going to assert this can we raise an error with a message here: "Reproducible mode is on, you must pass a batch index when calling this synth"

torchsynth/module.py
```diff
@@ -620,13 +620,18 @@ class Noise(SynthModule):
     these in a synth then choose different seeds for each instance.
     """
 
+    # Do we really want deterministic noise within each batch?
+    # https://github.com/turian/torchsynth/issues/250
```
jorshi, line 624: It's a good question -- it won't be perceptual

jorshi: But I can also see the argument for

torchsynth/module.py
turian: @jorshi make a separate PR. But keep in mind we're switching to batch_size 128 maybe

torchsynth/config.py
turian: will fix

torchsynth/module.py
turian: leave the comment for now

PR #287, opened 2021-04-09
Cpu vs gpu

turian:
** I DON'T THINK WE SHOULD MERGE THIS **

This unfortunately does not really change GPU vs CPU exp and log

Errors accumulate and are excerbated by pow operations

image


jorshi:
Interesting -- even running the pow operations with float64 we get this difference hey?

PR #288, opened 2021-04-10
Small memory optimizations

turian:
Seems to improve GPU memory usage according memlab
image

(BTW why do we have grads even in eval mode?? should we make an option for it?)

image


jorshi (approved):
Nice -- re the grad thing, I think all the param tensors still have the require_grad flag on even with torch.no_grad()? We could try having an option that loops through all the params and sets that flag to false to see what happens there.

PR #290, opened 2021-04-10
0.9.1

turian:
includes some previous PRs

### DIFF
--- a/torchsynth/module.py
+++ b/torchsynth/module.py
@@ -622,15 +622,15 @@ class Noise(SynthModule):
 
     # Do we really want deterministic noise within each batch?
     # https://github.com/turian/torchsynth/issues/250
-    noise_batch_size: int = 64
+    noise_batch_size: int = 128
     # Unfortunately, Final is not supported until Python 3.8
-    # noise_batch_size: Final[int] = 64
+    # noise_batch_size: Final[int] = 128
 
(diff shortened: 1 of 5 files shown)


torchsynth/module.py
```diff
@@ -622,15 +622,15 @@ class Noise(SynthModule):
...
     # Unfortunately, Final is not supported until Python 3.8
-    # noise_batch_size: Final[int] = 64
+    # noise_batch_size: Final[int] = 128
 
     def __init__(self, synthconfig: SynthConfig, seed: int, **kwargs):
         super().__init__(synthconfig, **kwargs)
 
         # https://github.com/turian/torchsynth/issues/255
-        assert self.batch_size == 64 or not self.synthconfig.reproducible
+        assert self.batch_size == 128 or not self.synthconfig.reproducible
```
jorshi, line 633: We've got more of these asserts around in config and synth that are still 64 -- can add a default batch size back in? even as a global in config or something so we don't have these asserts floating around

turian: @jorshi yes, make a PR. Put it in config.py and call it BATCH_SIZE_FOR_DETERMINISM

turian: https://github.com/turian/torchsynth/pull/292

PR #292, opened 2021-04-10
BATCH_SIZE_FOR_REPRODUCIBILITY

turian:


torchsynth/module.py
```diff
@@ -622,15 +622,18 @@ class Noise(SynthModule):
...
 
     def __init__(self, synthconfig: SynthConfig, seed: int, **kwargs):
         super().__init__(synthconfig, **kwargs)
 
         # https://github.com/turian/torchsynth/issues/255
-        assert self.batch_size == 64 or not self.synthconfig.reproducible
+        assert (
+            self.batch_size == BATCH_SIZE_FOR_REPRODUCIBILITY
+            or not self.synthconfig.reproducible
+        )
```
jorshi, line 636: I have a slight preference for raising an error with a message like we do in config here and below

jorshi (approved):

PR #308, opened 2021-04-11
Add SVG to README

turian:


turian:
Will do later when we have a longer README

PR #317, opened 2021-04-17
Modulation Mixer and Audio Mixer Param Names

jorshi:


turian:
Let wait until after optuna or this will be a pain for me.

PR #319, opened 2021-04-18
Load nebula

jorshi:
Loading and saving of hyperparams (nebula!)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
         "Source Code": "https://github.com/torchsynth/torchsynth/",
     },
     packages=find_packages(exclude=("tests", "examples")),
-    package_data={"torchsynth": ["nebulae/*.json"]},
+    package_data={"torchsynth": ["nebulae/voice/*.json"]},
     # package_dir={"": "src"},
     entry_points={"console_scripts": ["torchsynth.profile=torchsynth.profile:main"]},
     python_requires=">=3.7",
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
         "Source Code": "https://github.com/torchsynth/torchsynth/",
     },
     packages=find_packages(exclude=("tests", "examples")),
-    include_package_data=True,
+    package_data={"torchsynth": ["nebulae/*.json"]},
     # package_dir={"": "src"},
     entry_points={"console_scripts": ["torchsynth.profile=torchsynth.profile:main"]},
     python_requires=">=3.7",
--- a/setup.py
+++ b/setup.py
@@ -33,6 +33,7 @@ setup(
         "Source Code": "https://github.com/torchsynth/torchsynth/",
     },
     packages=find_packages(exclude=("tests", "examples")),
+    package_data={"": ["nebulae/*"]},
     # package_dir={"": "src"},
     entry_points={"console_scripts": ["torchsynth.profile=torchsynth.profile:main"]},
     python_requires=">=3.7",
--- a/setup.py
+++ b/setup.py
@@ -33,7 +33,7 @@ setup(
         "Source Code": "https://github.com/torchsynth/torchsynth/",
     },
     packages=find_packages(exclude=("tests", "examples")),
-    package_data={"": ["nebulae/*"]},
+    include_package_data=True,
     # package_dir={"": "src"},
     entry_points={"console_scripts": ["torchsynth.profile=torchsynth.profile:main"]},
     python_requires=">=3.7",
--- a/tests/test_synth.py
+++ b/tests/test_synth.py
@@ -2,7 +2,8 @@
 Tests for torch synths
 """
 
-
+import os
+import json
 import pytest
 import torch.nn
 import torch.tensor as tensor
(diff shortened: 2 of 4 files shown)


torchsynth/synth.py
```diff
@@ -327,6 +364,9 @@ def __init__(self, synthconfig: Optional[SynthConfig] = None, *args, **kwargs):
             ]
         )
 
+        # Load the nebula
+        self.load_hyperparameters(nebula)
```
jorshi, line 368: I kind of want this to be in AbstractSynth instead of in Voice. But, we don't want to force loading of hyperparams -- and I'm also not sure where it should be called since it needs to occur after all the modules and params are added.

torchsynth/synth.py
```diff
@@ -236,6 +239,34 @@ def set_hyperparameter(self, hyperparameter: Tuple[str, str, str], value: Any):
...
+    def load_hyperparameters(self, name: str) -> None:
+        """
+        Load hyperparameters from a JSON file
+        """
+
+        # Try to load nebulae from package resources, otherwise, try
+        # to load from a filename
+        try:
+            synth = type(self).__name__.lower()
+            nebulae_str = f"nebulae/{synth}/{name}.json"
```
jorshi, line 259: idea here is that we can add nebula to the package for each synth and that will exist in that package and you can just load it by calling the nebula name for that synth -- like "default" or "drum" and it will load that from the package resources. But, you can also just load nebula from a json file where ever. Doesn't have to be apart of the package.

tests/test_synth.py
```diff
@@ -130,8 +131,8 @@ def test_deterministic_noise(self):
             # cpu synth. Small numerical differences between computations of
             # GPU and CPU add up, so we need to relax the constraints here.
             # TODO https://github.com/torchsynth/torchsynth/issues/256
-            assert torch.mean(torch.abs(cuda11.detach().cpu() - x11)) < 2e-3
-            assert torch.mean(torch.abs(cuda2.detach().cpu() - x2)) < 3e-3
+            # assert torch.mean(torch.abs(cuda11.detach().cpu() - x11)) < 2e-1
+            # assert torch.mean(torch.abs(cuda2.detach().cpu() - x2)) < 2e-1
```
jorshi, line 135: These have been failing on my GPU :/ it's like 0.014 error now

torchsynth/synth.py
turian: Yeah nebulae are specific to specific synths. But I agree it could go into AbstractSynth somehow. For later: https://github.com/torchsynth/torchsynth/issues/321

setup.py
```diff
@@ -33,6 +33,7 @@
         "Source Code": "https://github.com/torchsynth/torchsynth/",
     },
     packages=find_packages(exclude=("tests", "examples")),
+    include_package_data=True,
```
turian, line 36: What is defined as package data? What gets rolled up into the pypi because of this?

jorshi: pretty much everything that is in the torchsynth package -- torchsynth/* 

I need to figure this one out though, I'd like to make it more specific. But it's also not working on travis.

setup.py
```diff
@@ -33,6 +33,7 @@
         "Source Code": "https://github.com/torchsynth/torchsynth/",
     },
     packages=find_packages(exclude=("tests", "examples")),
+    package_data={"torchsynth": ["nebulae/voice/*.json"]},
```
jorshi, line 36: this is better I think -- we will need to add the folders here for new synth nebulas when we add them (e.g. FM) -- but for now I think this is the cleanest way to add json files to the package

turian: nice

tests/test_synth.py
turian: Still think you should re-enable the test and make a comment about it.

torchsynth/nebulae/voice/default.json
```diff
@@ -0,0 +1,1250 @@
...
+ {
+  "name": [
+   "mixer",
+   "level0",
+   "symmetric"
+  ],
+  "value": false
+ },
+ {
+  "name": [
```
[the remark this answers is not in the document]
jorshi: just thinking forward to potentially having nested modules -- in which case it would be easier to keep these as just lists. If we have a nested module we just add another value. for example we could have `["fm_operator_1", "vco", "tuning", "curve"] and it would work with this format

torchsynth/nebulae/voice/default.json
```diff
@@ -0,0 +1,1250 @@
+[
+ {
```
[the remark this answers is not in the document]
jorshi: Sure. I agree it's a bit nasty to look at.

tests/test_synth.py
jorshi: Sure, will re-enable

torchsynth/nebulae/voice/default.json
```diff
@@ -0,0 +1,1250 @@
...
+  "value": 0.5
+ },
+ {
+  "name": [
+   "mod_matrix",
+   "level3_3",
+   "symmetric"
+  ],
+  "value": false
+ },
```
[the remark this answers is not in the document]
jorshi: done

torchsynth/nebulae/voice/default.json
```diff
@@ -0,0 +1,1250 @@
...
+  ],
+  "value": 1
+ },
+ {
+  "name": [
+   "vco_2",
+   "initial_phase",
+   "symmetric"
+  ],
+  "value": false
```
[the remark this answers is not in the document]
jorshi: in #320

torchsynth/synth.py
```diff
@@ -236,6 +239,34 @@ def set_hyperparameter(self, hyperparameter: Tuple[str, str, str], value: Any):
...
 
+    def save_hyperparameters(self, filename: str) -> None:
+        """
+        Save hyperparameters to a JSON file
+        """
+        hp = [{"name": key, "value": val} for key, val in self.hyperparameters.items()]
+        with open(os.path.abspath(filename), "w") as fp:
+            json.dump(hp, fp, indent=True)
+
+    def load_hyperparameters(self, name: str) -> None:
```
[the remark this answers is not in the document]
jorshi: I kind of like the non-tricky approach suggestion -- probably the easier to understand from an API perspective?

torchsynth/nebulae/voice/default.json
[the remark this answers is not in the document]
turian: okay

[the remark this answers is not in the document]
turian: I don't think this will be the right approach for nesting but we can keep it for now

PR #325, opened 2021-04-19
Nebulae cleanup

jorshi:


tests/test_synth.py
```diff
@@ -131,8 +131,8 @@ def test_deterministic_noise(self):
             # cpu synth. Small numerical differences between computations of
             # GPU and CPU add up, so we need to relax the constraints here.
             # TODO https://github.com/torchsynth/torchsynth/issues/256
-            # assert torch.mean(torch.abs(cuda11.detach().cpu() - x11)) < 2e-1
-            # assert torch.mean(torch.abs(cuda2.detach().cpu() - x2)) < 2e-1
+            assert torch.mean(torch.abs(cuda11.detach().cpu() - x11)) < 2e-1
+            assert torch.mean(torch.abs(cuda2.detach().cpu() - x2)) < 2e-1
```
jorshi, line 135: Re-enabling these -- issue is already linked above

torchsynth/synth.py
```diff
@@ -243,24 +244,35 @@ def save_hyperparameters(self, filename: str) -> None:
...
         hp = [{"name": key, "value": val} for key, val in self.hyperparameters.items()]
+        hp = json.dumps(hp, indent=True)
+        hp = util.format_hyperparameter_json(hp)
+
         with open(os.path.abspath(filename), "w") as fp:
-            json.dump(hp, fp, indent=True)
+            fp.write(hp)
 
-    def load_hyperparameters(self, name: str) -> None:
+    def load_hyperparameters(self, nebula: str) -> None:
```
jorshi, line 255: Decided to keep this as is. If we add multiple args here for the name vs. filename then we would need to have that in the Voice constructor as well. Unless we decide that we only support loading of torchsynth package nebulae from a synth constructor.

Also. Do we want to call these functions `load_nebula` and `save_nebula`?

turian: Why not? Do it in the next PR I guess :)

Maybe that means we also have to change hyperparameters in synth to nebula?

torchsynth/util.py
```diff
@@ -50,6 +52,29 @@ def normalize(signal: Signal) -> Signal:
...
+          "curve"
+        ],
+
+    Becomes:
+        "name": ["adsr_1", "attack", "curve"],
+    """
+    hp = re.sub(
+        r'("name": \[)(\n\s+)(.*)(\n\s+)(.*)(\n\s+)(.*)(\n\s+)',
+        r"\1\3 \5 \7",
+        hyperparameter_json,
```
turian, line 73: Oh god I think this a bit scary. If you're going to do this by hand, then in this function add a proper assert:

```
assert hyperparameter_json == json.loads(hp), "Your hyperparameters cannot easily be converted to JSON using our hacky code"
```

or some such. Like what if some nerd decide to add quotes or commas to their param names?

turian: or maybe a try except and then if except log the message and say we're using ugly verbose standard JSON

torchsynth/synth.py
jorshi: Will leave for a future PR -- made an issue for it #330

torchsynth/util.py
jorshi: Yah you know what I'm not sure how necessary it is to have this in the library -- if we want we could create a little util script or something to convert json files to a more clean format. Or have an arg that applies this if we think it would be helpful. But I'd be just as happy to ditch this.

turian: Let's ditch it and do it by hand when we commit to the repo?

turian (approved):

PR #327, opened 2021-04-20
Reproducible batch size

jorshi:
Reproducibility with batch sizes that are multiples of 32

### DIFF
--- a/tests/test_module.py
+++ b/tests/test_module.py
@@ -8,7 +8,7 @@ import torch.tensor as tensor
 
 import torchsynth
 import torchsynth.module as synthmodule
-from torchsynth.config import BATCH_SIZE_FOR_REPRODUCIBILITY, SynthConfig
+from torchsynth.config import BASE_REPRODUCIBLE_BATCH_SIZE, SynthConfig
 
 
 class TestSynthModule:
@@ -195,9 +195,9 @@ class TestSynthModule:
         # of noise signals. All these noise samples should equal each other.
         # i.e., noise should be returned deterministically regardless of the
         # batch size.
-        synthconfig32 = SynthConfig(32, reproducible=False)
-        synthconfig64 = SynthConfig(64, reproducible=False)
-        synthconfig128 = SynthConfig(128, reproducible=False)
+        synthconfig32 = SynthConfig(32)
+        synthconfig64 = SynthConfig(64)
+        synthconfig128 = SynthConfig(128)
 
         noise32 = synthmodule.Noise(synthconfig32, seed=0)
         noise64 = synthmodule.Noise(synthconfig64, seed=0)
@@ -215,11 +215,9 @@ class TestSynthModule:
         assert torch.all(out3 != out4)
 
         with pytest.raises(ValueError):
-            # If the batch size >= BATCH_SIZE_FOR_REPRODUCIBILITY
-            # but not divisible by it, it should raise an error
-            synthconfigwrong = SynthConfig(
-                BATCH_SIZE_FOR_REPRODUCIBILITY + 1, reproducible=False
-            )
+            # If the batch size is not a multiple of BASE_REPRODUCIBLE_BATCH_SIZE,
+            # it should raise an error in reproducible mode
+            synthconfigwrong = SynthConfig(BASE_REPRODUCIBLE_BATCH_SIZE + 1)
             synthmodule.Noise(synthconfigwrong, seed=0)
 
 
--- a/tests/test_reproducibility.py
+++ b/tests/test_reproducibility.py
@@ -11,7 +11,6 @@ from torchsynth.synth import Voice
 
 class TestReproducibility:
     def test_voice_reproducibility(self):
-        # TODO make this work with different batch sizes
         synthconfig = SynthConfig()
         voice_1 = Voice(synthconfig)
         voice_2 = Voice(synthconfig)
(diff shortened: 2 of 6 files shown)


tests/test_reproducibility.py
```diff
@@ -36,7 +35,39 @@ def test_voice_reproducibility(self):
...
+                voice32(7),
+            ]
+        )
+
+        # TODO there are some unexpected, very small numerical
+        # errors between some, but not all, batch sizes
+        # See https://github.com/torchsynth/torchsynth/issues/326
+        assert torch.all(torch.isclose(out256, out128))
+        assert torch.all(torch.isclose(out256, out64))
+        assert torch.all(torch.isclose(out256, out32))
```
jorshi, line 70: This was weird. I was initially checking to see if they were equal and out256 == out64 and out256 == out32, but, out256 != out128. Looked into it and there are 40 samples all with an error of 7.450581e-09 at about 3min 40s into synth1B1-182. But the error only exists between batch size of 128 and the others. I checked the parameters and the noise and they are exactly the same. This one is boggling me a bit lol.

torchsynth/synth.py
```diff
@@ -272,19 +272,6 @@ def randomize(self, seed: Optional[int] = None):
...
-            self.batch_size != BATCH_SIZE_FOR_REPRODUCIBILITY
-            and self.synthconfig.reproducible
-        ):
-            raise ValueError(
-                "Reproducibility currently only supported "
-                f"with batch_size = {BATCH_SIZE_FOR_REPRODUCIBILITY}. "
-                "If you want a different batch_size, "
-                "initialize SynthConfig with reproducible=False"
-            )
-
```
jorshi, line 287: The code below already supports any batch size.

torchsynth/config.py
```diff
@@ -7,7 +7,8 @@
 # and abstract synth parameter randomization
 # (https://github.com/torchsynth/torchsynth/issues/253)
 # are non-reproducible unless batch_size == BATCH_SIZE_FOR_REPRODUCIBILITY.
-BATCH_SIZE_FOR_REPRODUCIBILITY = 128
```
turian, line 10: Docs need to be updated to explain what each of these mean and the comment above

tests/test_reproducibility.py
turian: On GPUs? They're not deterministic at low epsilons. The comment is good

torchsynth/config.py
```diff
@@ -62,15 +63,12 @@ def __init__(
...
         if self.reproducible:
-            # Currently, noise module
-            # (https://github.com/torchsynth/torchsynth/issues/255)
-            # and abstract synth parameter randomization
-            # (https://github.com/torchsynth/torchsynth/issues/253)
-            # are non-reproducible unless batch_size == BATCH_SIZE_FOR_REPRODUCIBILITY.
-            if batch_size != BATCH_SIZE_FOR_REPRODUCIBILITY:
+            # We currently only support reproducibility with batch sizes that
+            # are multiples of BASE_REPRODUCIBLE_BATCH_SIZE
+            if batch_size % BASE_REPRODUCIBLE_BATCH_SIZE != 0.0:
```
turian, line 68: wait why not 0 not 0.0 since these are longs or ints?

torchsynth/module.py
```diff
@@ -622,62 +622,39 @@ class Noise(SynthModule):
 
     # Do we really want deterministic noise within each batch?
```
turian, line 623: Maybe change this comment to explain the reasoning

torchsynth/module.py
```diff
@@ -622,62 +622,39 @@ class Noise(SynthModule):
...
+                    f"Batch size must be a multiple of {self.noise_batch_size} "
+                    "when using reproducible mode. Either change your batch size,"
+                    "or set reproducible=False in the SynthConfig for this module."
                 )
 
-            noise = noise.repeat(self.batch_size // self.noise_batch_size, 1)
+            noise = torch.empty((self.noise_batch_size, self.buffer_size), device="cpu")
+            noise.data.uniform_(-1.0, 1.0, generator=generator)
+            if self.batch_size > self.noise_batch_size:
+                noise = noise.repeat(self.batch_size // self.noise_batch_size, 1)
```
turian, line 648: I'm a bit scared of this lol, did you write a test? Maybe it's later in the PR review. I haven't used .repeat much

torchsynth/synth.py
turian: Yeah looks good. Have you profiled? Does this PR affect speed at all

torchsynth/module.py
turian: Don't see a test can we write one specifically for noise? I know we have one for synth but it would be nice

tests/test_reproducibility.py
jorshi: Ya I'll extend this test to check on GPUs if they're available

torchsynth/synth.py
jorshi: The code is already in main and is included in the profiles we did for the paper :)

jorshi: i.e. the solution we came up with a week ago or whatever was deterministic for different batch sizes and is also fast.

torchsynth/config.py
jorshi: Created an issue for it: #331

torchsynth/config.py
jorshi: Good catch -- will fix

tests/test_module.py
```diff
@@ -195,9 +195,9 @@ def test_noise(self):
         # of noise signals. All these noise samples should equal each other.
         # i.e., noise should be returned deterministically regardless of the
         # batch size.
-        synthconfig32 = SynthConfig(32, reproducible=False)
-        synthconfig64 = SynthConfig(64, reproducible=False)
-        synthconfig128 = SynthConfig(128, reproducible=False)
+        synthconfig32 = SynthConfig(32)
+        synthconfig64 = SynthConfig(64)
+        synthconfig128 = SynthConfig(128)
```
jorshi, line 200: this test confirms reproducibility for different batch sizes of noise

torchsynth/module.py
jorshi: Yup we have one -- made a comment above to point it out

jorshi: Already had one that tested noise reproducibility -- updated it slightly in this PR to make sure reproducible mode is on

PR #333, opened 2021-04-22
Module Documentation

jorshi:
This is a start on getting synth modules to show up nicely in the documentation


docs/source/api/module.md
```diff
@@ -2,8 +2,146 @@ torchsynth.module
 =================
 
 ```{eval-rst}
-.. automodule:: torchsynth.module
-   :members:
-   :undoc-members:
-   :show-inheritance:
+SynthModule
```
jorshi, line 5: This is way more verbose I realize -- and we will need to update this every time we add a module. But this allows the theme to create a table of contents with modules grouped by audio rate, control rate, and parameter modules which I really like.

jorshi: There may be a way to template this out in the future. But I think this works for now.

docs/source/conf.py
```diff
@@ -151,3 +159,52 @@
 html_css_files = [
     "css/custom.css",
 ]
+
+
+class TorchSynthParameterDoc(ClassDocumenter):
```
jorshi, line 164: The module parameter ranges were printing really gross -- this targets attributes that are lists of ModuleParameterRanges and iterates through and prints them nicely in a list with the range / default hyperparams

setup.py
```diff
@@ -70,6 +70,7 @@
             # Temporarily disabled so we can push to pypi
             # "pt-lightning-sphinx-theme @ https://github.com/PyTorchLightning/lightning_sphinx_theme/tarball/master#egg=pt-lightning-sphinx-theme",
             "sphinxcontrib-napoleon",
+            "sphinx-autodoc-typehints",
```
jorshi, line 73: Automatically infer types from annotations

torchsynth/module.py
```diff
@@ -151,7 +155,7 @@ def get_parameter(self, parameter_id: str) -> ModuleParameter:
         Get a single ModuleParameter for this module
 
         Args:
-            parameter_id: str: Id of the parameter to return
```
jorshi, line 154: we don't need to specify types in the docstring with the typehints package

torchsynth/module.py
```diff
@@ -363,11 +398,10 @@ def _ramp(
         method works by tilting and clipping ramps between 0 and 1, then
         applying some scaling factor (`alpha`).
 
-        Parameters
-        ----------
-        duration T: Length of the ramp in seconds
-        start Optional[T]: Initial delay of ramp in seconds. Defaults to None (0 start)
```
jorshi, line 369: Also don't need to specify default values -- this shows up in the method template in the docs

docs/source/api/module.md
```diff
@@ -2,8 +2,146 @@ torchsynth.module
...
+    :undoc-members:
+    :show-inheritance:
+
+VCO
+===
+.. autoclass:: torchsynth.module.VCO
+    :members:
+    :undoc-members:
+    :show-inheritance:
+    :private-members: _forward
```
jorshi, line 72: One thing I'm not sure about yet is how to document the forward for each of our synthmodules. Since the args are defined in `_forward` -- but we don't want users to call `_forward` directly ever. If anything they should use `__call__`

turian: @jorshi rename it to `__forward` and then it can't really be called externally?

turian: IDK, maybe we need to have a quick chat?

docs/source/conf.py
turian: Wow

PR #348, opened 2021-04-25
Changes

Doddox-UV:
Batch-processing.md images size fixed
torchsynth.parameter.py updated to fix the error


turian:
Breaks our tests

PR #362, opened 2021-05-05
Pin sklearn version

turian:
Based upon https://github.com/torchsynth/torchsynth/pull/361

(Is sklearn a dependency of our main dependencies? Or just our dev ones??)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -61,6 +61,7 @@ setup(
             "pytest-cov",
             "ipython",
             "librosa",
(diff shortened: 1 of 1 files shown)


jorshi:
took a quick look and it appears like sklearn is only a dependency in dev -- I think librosa depends on it

turian:
@jorshi snyk looks at whatever we put in requirements.txt and ignores setup.py. I think it's fine to audit dev libraries too. Our install instructions are all setup.py based anyhow

PR #367, opened 2021-06-10
Time-varying FIR Filter

jorshi:
Builds off code that @maxsolomonhenry started for this and implements a time-varying lowpass FIR using overlap add fft convolution.


torchsynth/filter.py
```diff
@@ -0,0 +1,252 @@
...
+
+class SincFilterBase(TimeVaryingFIRBase):
+    """
+    Implements filters using the the windowed sinc method to generate impulses.
+
+    TODO: Would be nice to have a cutoff and mod_depth parameter -- and then have
+        the cutoff control signal be optional (like a filter module that is optionally
+        modulate-able). Also, maybe have the modulation CV be 0-1 range (which is what
+        most of our control signals operate at) -- and then scale that to log frequency
+        in a range from 0 - nyquist.
```
jorshi, line 191: @maxsolomonhenry ?

torchsynth/filter.py
```diff
@@ -0,0 +1,252 @@
+"""
+Time-varying filters.
+"""
```
jorshi, line 3: What do you guys think about having this in a separate module like this? I was getting a weird sphinx error when trying to generate docs for this that wasn't occurring when it was in module, so that is weird ... need to look into that a bit more

turian: It's fine by me

examples/examples.py
```diff
@@ -59,9 +59,8 @@ def display(*args, **kwargs):
 
 import numpy as np
 import numpy.random
-import torch.fft
-import torch.tensor as tensor
-from torch import Tensor as T
+import torch
+from torch import tensor
```
jorshi, line 63: Update of tensor import for torch 1.9

torchsynth/filter.py
maxsolomonhenry: Sounds reasonable!

torchsynth/filter.py
maxsolomonhenry: +1

maxsolomonhenry:
(Caveat, I'd love to run this code, been struggling all weekend with my M1 getting a development environment off the ground. Nothing but headaches. Leaving this comment here to keep things moving forward...)

General note, re: _delay compensation_ in the FIR base class... 

I would move this to `get_impulse_matrix` as the delay is specific to the kernel. The above assumes a symmetrical IR (which would then also be linear phase). This might be a good default assumption, but it doesn't make sense for the generic base class to impose this delay compensation.

I would suggest some sort of "delay reporting" method to be implemented in the child class, and this could optionally be folded into the output method somehow. (i.e., make the above `_output` and generic to all classes, and make the child implement `output` method, which optional delay compensation).

jofrevalles:
Hey there! Is this brach currently working?

jorshi:
@jofrevalles -- last I tested the FIR code was working correctly! But that was quite a while ago. Unfortunately we all have moved onto other projects and haven't had the time to maintain torchsynth -- But, If you checkout this branch and merge main into it, it *should* work. Let me know!

jofrevalles:
> @jofrevalles -- last I tested the FIR code was working correctly! But that was quite a while ago -- If you checkout this branch and merge main into it, it _should_ work. Let me know!

Thank you, @jorshi ! I will use the branch and let you know if it works!

PR #368, opened 2021-06-18
Update tensor imports for torch 1.9

jorshi:


turian:
0) Do you remember why we did this originally? Because <= 1.8 Tensor is the type for typing?

1) If we want to make sure this works for older torch versions, can you a) create a fork of https://github.com/torchsynth/torchsynth/pull/280/files called travis-1.9 and merge this into it and b) update the travis matrix so that every configuration where we have torch-1.8 we also try 1.9 now?

Just want to make sure we don't break stuff backwards.

2) We should have a placeholder 1.0.1 pin that we will merge and push to pypi later.

jorshi:
@turian 

0. I don't recall -- we originally had `import torch.tensor as T` - but torch.tensor is not the type, so we changed it to `from torch import Tensor as T` and then changed `import torch.tensor as tensor` -- but the imports have changed in the 1.9.

This is not the direct commit that changed things, but is a useful message:
https://github.com/pytorch/pytorch/commit/b0afe945a7770ca64da9b8e1065ce0f572dedd49#diff-41eb33210e32f7a565a420a78573740a643b5f124726e7013e8b96aad158948e
> This BC-breaking only for users that rely on unintended behavior. Since `torch/__init__.py` loaded `torch/tensor.py` it was populated in `sys.modules`. `torch/__init__.py` then overwrote `torch.tensor` with the actual function. With this `import torch.tensor as tensor` does not fail, but returns the function rather than the module. Users that rely on this import need to change it to `from torch import tensor`.

1. Yes would be good to clean up travis, I created a new PR #369 with torch 1.9 as suggested and ran https://travis-ci.com/github/torchsynth/torchsynth/builds/230419972 -- looks like this change is okay for torch 1.8. We are still breaking on torch < 1.8, i looked into a bit and it is more complicated than just swapaxes. So I think we just drop those - unless we really want to spend the time digging into it. There are other configs that can be dropped as well. Side note on travis is that we have used up all our credits (probably on these big runs). 

2. Yes

Issue #377, opened 2022-08-15
Cannot use large batch index


turian:
Large batch indexes are broken for reasons that are not documented or easy to understand.

[Colab to replicate](https://colab.research.google.com/drive/1Sn4u95SvV-fjyPWNYQDpXm4CZvGXC5Kr?usp=sharing)

Code:
```
!pip3 install torchsynth

from torchsynth.config import SynthConfig
from torchsynth.synth import Voice
import torch

# We'll generate BATCH_SIZE sounds per batch, 4 seconds each
# TODO: On larger GPUs, use larger batch size
synthconfig = SynthConfig(
    batch_size=256,
    reproducible=False,
    sample_rate=44100,
    buffer_size_seconds=4.0,
)

voice = Voice(synthconfig=synthconfig)

# Run on the GPU if it's available
if torch.cuda.is_available():
    voice = voice.to("cuda")

from tqdm.auto import tqdm

# This works
for i in tqdm(list(range(100))):
  voice(i)

# This doesnt???
for i in tqdm(list(range(1024 * 1024 * 1024, 1024 * 1024 * 1024 + 100))):
  voice(i)
```

Here's the error:
```
---------------------------------------------------------------------------
AssertionError                            Traceback (most recent call last)
[<ipython-input-6-cc1b9a3cd814>](https://localhost:8080/#) in <module>
      1 for i in tqdm(list(range(1024 * 1024 * 1024, 1024 * 1024 * 1024 + 100))):
----> 2   voice(i)

2 frames
[/usr/local/lib/python3.7/dist-packages/torch/nn/modules/module.py](https://localhost:8080/#) in _call_impl(self, *input, **kwargs)
   1128         if not (self._backward_hooks or self._forward_hooks or self._forward_pre_hooks or _global_backward_hooks
   1129                 or _global_forward_hooks or _global_forward_pre_hooks):
-> 1130             return forward_call(*input, **kwargs)
   1131         # Do not call functions when jit is used
   1132         full_backward_hooks, non_full_backward_hooks = [], []

[/usr/local/lib/python3.7/dist-packages/torchsynth/synth.py](https://localhost:8080/#) in forward(self, batch_idx, *args, **kwargs)
    220                 "pass a batch index when calling this synth"
    221             )
--> 222         is_train = self._batch_idx_to_is_train(batch_idx)
    223         if self.synthconfig.no_grad:
    224             with torch.no_grad():

[...]

PR #378, opened 2022-08-16
Updating torch.range to torch.arange

jorshi:
[`torch.range` is now deprecated](https://pytorch.org/docs/stable/generated/torch.range.html) and should be replaced with `torch.arange`, which is consistent with pythons built-in range. `torch.range` also was producing errors with high-valued ranges, see #377

Merge #380 into this first.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -41,7 +41,7 @@ setup(
         "numpy",
         "scipy",
         "torch>=1.8",
-        "pytorch-lightning",
+        "pytorch-lightning>=1.4",
         # pypi release (only master) doesn't support OrderedDict typing
         # "typing-extensions",
     ],
--- a/torchsynth/profile.py
+++ b/torchsynth/profile.py
@@ -110,7 +110,7 @@ def run_lightning_module(
         # Run module with profiling
         pr = cProfile.Profile()
         pr.enable()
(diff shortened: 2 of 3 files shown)


turian:
Yeah but what about older versions of pytorch? Do we break them? We should run a full jenkins

turian:
Since when is it deprecated? Do you have a link?

jorshi:
@turian `torch.range` has been deprecated in favour of `torch.arange` since torch v0.1 https://github.com/pytorch/pytorch/releases/tag/v0.1.12 -- so we should be good for backward compatibility :)

turian:
Okay. Why are tests still breaking?

jorshi:
Tests were breaking because the deprecated function call `Trainer.test(test_dataloader=)` was removed from Lightning (formally PTL) in [version 1.6](https://github.com/Lightning-AI/lightning/releases/tag/1.6.0), see https://github.com/Lightning-AI/lightning/pull/10325. `test_dataloader` was deprecated in [v1.4.0](https://github.com/Lightning-AI/lightning/releases/tag/1.4.0), see https://github.com/Lightning-AI/lightning/pull/7431. Replacing `Trainer.test(test_dataloader=)` with `Trainer.test(dataloader=)` will work for ptl>=1.4

PR #380, opened 2022-08-17
Codecov Action Fix

jorshi:
In more recent versions of torch it appears that a temporary file `_remote_module_non_scriptable.py` is created, which breaks codecov xml generation. This change omits that file during xml generation,


turian:
uh wow

PR #384, opened 2022-08-19
Add new_empty to Signal

turian:
Some of the trivial changes are because of newer versions of black.

The new method `Signal.new_empty` and its test enable us to deepcopy Signals and thus checkpoint them.

### DIFF
--- a/examples/simplesynth.py
+++ b/examples/simplesynth.py
@@ -11,19 +11,18 @@
 # +
 from typing import Optional
 
-import torch
 import IPython.display as ipd
+import torch
 
+from torchsynth.config import BASE_REPRODUCIBLE_BATCH_SIZE, SynthConfig
 from torchsynth.module import (
     ADSR,
+    VCA,
     ControlRateUpsample,
     MonophonicKeyboard,
     SquareSawVCO,
-    VCA,
 )
 from torchsynth.synth import AbstractSynth
-from torchsynth.config import SynthConfig, BASE_REPRODUCIBLE_BATCH_SIZE
-
 
 # -
 
--- a/tests/test_signal.py
+++ b/tests/test_signal.py
@@ -1,0 +1,21 @@
+"""
+Tests for torch signals
+"""
+
+from copy import deepcopy
+
+import torch
+
+from torchsynth.signal import Signal
+
+
+class TestSignal:
+    """
+    Tests for Signal
+    """
+
+    device = "cuda" if torch.cuda.is_available() else "cpu"
+
+    def test_deepcopy(self):
+        signal = torch.zeros(1, 1).as_subclass(Signal)
+        print(deepcopy(signal))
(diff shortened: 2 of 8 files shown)


torchsynth/signal.py
```diff
@@ -20,3 +20,9 @@ def batch_size(self):
     def num_samples(self):
         assert self.ndim == 2
         return self.shape[1]
+
+    def new_empty(self, *args, **kwargs):
+        """
+        Implement [torch.Tensor.new_empty](https://pytorch.org/docs/stable/generated/torch.Tensor.new_empty.html) so that Signal can be [`deepcopy`](https://docs.python.org/3/library/copy.html#copy.deepcopy)ed.
```
jorshi: This comment is triggering lint on line-too-long

examples/simplesynth.py
```diff
@@ -11,19 +11,18 @@
...
 from typing import Optional
 
-import torch
 import IPython.display as ipd
+import torch
 
+from torchsynth.config import BASE_REPRODUCIBLE_BATCH_SIZE, SynthConfig
 from torchsynth.module import (
     ADSR,
+    VCA,
```
jorshi, line 20: This was in alphabetically order?

tests/test_signal.py
```diff
@@ -0,0 +1,21 @@
...
+class TestSignal:
+    """
+    Tests for Signal
+    """
+
+    device = "cuda" if torch.cuda.is_available() else "cpu"
+
+    def test_deepcopy(self):
+        signal = torch.zeros(1, 1).as_subclass(Signal)
+        print(deepcopy(signal))
```
jorshi, line 21: does deepcopy calls `torch.new_empty()`?

jorshi (approved):

torchsynth/signal.py
turian: It's fixed now

examples/simplesynth.py
turian: idk I just upgraded to latest black and isort and got these results

tests/test_signal.py
turian: Yup

torchsynth/signal.py
```diff
@@ -20,3 +20,15 @@ def batch_size(self):
     def num_samples(self):
         assert self.ndim == 2
         return self.shape[1]
+
+    def new_empty(self, *args, **kwargs):
```
jorshi, line 24: Actually one thought -- the torch tensor version https://pytorch.org/docs/stable/generated/torch.Tensor.new_empty.html  has a positional arg `size` -- should this also have that as opposed to just `*args`?

turian: nah

*args is easier and more flexible for different torch version

turian: more agnostic this way

jorshi: Or does specifying it as `*args` provide a bit more flexibility and future proof wrapping this?

PR #394, opened 2022-10-14
Getting and setting parameter values

jorshi:
Adding in a couple methods in AbstractSynth to make it easier to get and set all parameters at once using tensors.

### DIFF
--- a/torchsynth/synth.py
+++ b/torchsynth/synth.py
@@ -148,27 +148,18 @@ class AbstractSynth(LightningModule):
 
         return OrderedDict(parameters)
 
-    def get_raw_parameter_values(self, include_frozen=True):
+    def get_raw_parameter_values(self):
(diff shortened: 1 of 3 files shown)


torchsynth/synth.py
```diff
@@ -165,6 +178,25 @@ def set_parameters(
...
+    def set_raw_parameter_values(self, x):
+        """
+        Directly set the values for all synth parameters. Expects a 2D `torch.Tensor`
+        with a shape of (batch_size, num_parameters) containing raw parameter
+        values in the range [0,1].
+
+        Args:
+            x: `torch.Tensor` containing new parameter values. Expected shape is
+                (batch_size, num_parameters).
+        """
```
jorshi, line 190: Assert value range is [0,1]

PR #423, opened 2023-11-13
Sphinx Builds

jorshi:
- Creating a separate set of dependencies for building the Sphinx docs. Jupytext and document dependencies don't play nice.
- Bumping the minimum python version up to 3.8 to support minimum PTL requirement
- Updated the workflow for verifying doc builds work

This just fixes the test on documentation building. Will need address documentation deployment separately #424


jorshi:
Sphinx action doesn't seem to be supported any more. Looks like we can probably update this though: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-static-site-generators-with-python

turian (approved):

PR #469, opened 2024-10-07
Add AbstractSynth.set_parameters_0to1

turian:
This doesn't break any existing functionality, so I think it's safe to merge.


jorshi:
Looks okay to me -- this allows you to set parameters for an entire synth passing in values in a 0-1 range?

